PR #34, opened 2023-07-15
Better error handling

phil-scale:


clients/python/llmengine/data_types.py
```diff
@@ -183,7 +183,7 @@ class CompletionSyncV1Request(BaseModel):
 
     prompts: List[str] = Field(..., min_items=1)
     max_new_tokens: int = Field(..., gt=0)
-    temperature: float = Field(..., ge=0.0)
+    temperature: float = Field(..., gt=0.0)
```
yixu34, line 186: I think we need to do another copy from the server DTOs? Because @yunfeng-scale has it at `(0, 100]` here: https://github.com/scaleapi/models/pull/7128/files#diff-f81fae32377eba9c6da15635ede3effeee67fea20e1426addd20b0e72679572cR115

yixu34:
Nice

PR #41, opened 2023-07-16
Fix CreateFineTuneJobRequest DTO

squeakymouse:
Since it's changing in Spellbook

### DIFF
--- a/clients/python/llmengine/__init__.py
+++ b/clients/python/llmengine/__init__.py
@@ -26,7 +26,7 @@ from llmengine.data_types import (
     CreateFineTuneRequest,
     CreateFineTuneResponse,
     GetFineTuneResponse,
-    ListFineTuneResponse,
+    ListFineTunesResponse,
     TaskStatus,
 )
 from llmengine.fine_tuning import FineTune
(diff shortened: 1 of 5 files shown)


clients/python/llmengine/data_types.py
```diff
@@ -255,18 +255,16 @@ class CompletionStreamV1Response(BaseModel):
 
 
 class CreateFineTuneJobRequest(BaseModel):
+    model: str
```
yixu34, line 258: Curious, did you make these changes by hand, or did you copy from the server? I think @phil-scale tried to set up a script to do that but it's not in place yet.

In any case, we should manually copy from the server for now.

Also, does this PR attempt to solve https://app.shortcut.com/scaleai/story/782421/finetunejob-should-be-renamed-to-finetune-in-documentation?

clients/python/llmengine/__init__.py
```diff
@@ -18,32 +18,32 @@
...
     CompletionStreamV1Response,
     CompletionSyncV1Response,
-    CreateFineTuneJobRequest,
-    CreateFineTuneJobResponse,
-    GetFineTuneJobResponse,
-    ListFineTuneJobResponse,
+    CreateFineTuneRequest,
+    CreateFineTuneResponse,
+    GetFineTuneResponse,
+    ListFineTuneResponse,
```
yixu34, line 29: Similar to https://github.com/scaleapi/scaleapi/pull/63339#discussion_r1264749098

yixu34 (approved):

PR #46, opened 2023-07-17
Update completions.md

luvkothari:
Added more details around methods

### DIFF
--- a/docs/guides/completions.md
+++ b/docs/guides/completions.md
@@ -1,3 +1,42 @@
-Given a prompt, the model will return one or more predicted Completions. The Completions API can be run either
-synchronous or asynchronously (via Python `asyncio`); for each of these modes, you can also choose to stream token
-responses or not.
+# Overview
+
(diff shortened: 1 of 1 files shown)


docs/guides/completions.md
```diff
@@ -1,3 +1,42 @@
...
+2. **acreate()**: Asynchronous method for creating a completion task.
+
+Both methods return either a **CompletionSyncV1Response** object or a stream of **CompletionStreamV1Response** objects, depending on the stream parameter.
+
+# Method Details
+**1. Create()**
+
+This method is used to create a completion task synchronously.
+
+Parameters:
```
yixu34, line 17: I don't think we want to duplicate the detailed API reference in here. Otherwise we run the risk of things getting out of date.

Maybe we want to treat this as more of a tutorial/cookbook that's based heavily on examples.

yixu34: https://platform.openai.com/docs/guides/gpt/completions-api

I might just add some examples, and add a "which model should I use?"

yixu34:
Closing, out of date now

PR #49, opened 2023-07-17
Flesh out the docstrings for completion input params

yixu34:
@seanshi-scale @squeakymouse I think y'all will want to do something similar for fine tuning.

https://app.shortcut.com/scaleai/story/782480/documentation-for-completions-should-contain-more-details

### DIFF
--- a/clients/python/llmengine/data_types.py
+++ b/clients/python/llmengine/data_types.py
@@ -187,8 +187,12 @@ class CompletionSyncV1Request(BaseModel):
 
 
 class CompletionOutput(BaseModel):
+    """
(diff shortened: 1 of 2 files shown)


clients/python/llmengine/data_types.py
```diff
@@ -187,8 +187,12 @@ class CompletionSyncV1Request(BaseModel):
 
 
 class CompletionOutput(BaseModel):
+    """
+    Represents the output of a completion request to a model.
+    """
+
     text: str
-    """Text"""
```
seanshi-scale, line 191: lmao

seanshi-scale (approved):

PR #50, opened 2023-07-17
fine-tuning cookbook

seanshi-scale:
cookbook for finetuning

val acc is 0.94 🤷


rkaplan:
Quick feedback:
- Missing how to install scikit-learn so people can run the example themselves without import errors
- Please change this section to follow Tiger's example for how upload a public CSV to pastebin, so users can follow along themselves if they don't have S3 set up:
image
- Document how long you expect the fine-tuning process to take
- Please add newlines to make the Completion.create call on this line fit in one line:
image
- Nit: stylistically, better to use a `for / else` block in Python than ending the method with a top level `return "Error"`
- Please add a few more sentences to the very beginning about what we are trying to do and why fine tuning makes sense for it. You can borrow from https://scaleapi.github.io/llm-engine/guides/fine_tuning/

yixu34 (approved):

PR #55, opened 2023-07-17
Fixes to docs:

st-scale:
- Much richer 'Completions' how-to guide
- document 'TaskStatus', remove documentation of 'traceback' in completion status
- examples using an instruct model for friendlier outputs
- remove empty langchain section from main page

### DIFF
--- a/clients/python/llmengine/data_types.py
+++ b/clients/python/llmengine/data_types.py
@@ -109,6 +109,12 @@ class GetModelEndpointV1Response(BaseModel):
 
 
 class TaskStatus(str, Enum):
+    """
+    Specifies the status of a request as an enumeration.
+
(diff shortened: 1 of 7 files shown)


clients/python/llmengine/data_types.py
```diff
@@ -109,6 +109,12 @@ class GetModelEndpointV1Response(BaseModel):
 
 
 class TaskStatus(str, Enum):
+    """
```
yixu34, line 112: We're actually going to remove this from the API, don't think it was supposed to be there.

st-scale: Removing it from both CompletionSyncV1Response and CompletionStreamV1Response ? I can modify the docs assuming that's true. All error codes will be communicated via HTTP response codes?

yixu34: Yeah, this is removed in #57

PR #57, opened 2023-07-17
Update `Completions` response fields

phil-scale:
Adds the following updates:
* Removes `status`, `traceback`, and `num_prompt_tokens` from response
* Adds `request_id` to response
* Uses `model` instead of `model_name` when creating a completion


yixu34:
Oh oops didn't see this PR cc @yunfeng-scale

yixu34 (approved):

PR #59, opened 2023-07-17
Apply Arun's completion guides patch

yunfeng-scale:


yunfeng-scale:
don't think this PR would cause integration test failure

yunfeng-scale:
we're not installing istio into minikube. the failure is expected

PR #82, opened 2023-07-17
Update finetuning docs

felixs8696:
Address:
The output should contain a fine_tuned_model like OpenAI. We should also update the docs to reflect that this fine_tuned_model is what gets sent to Completions.

We should also document how the suffixing works: f"{model}:{suffix}:{time_now}" if suffix is not None else f"{model}.{time_now}"

[[sc-782703]](https://app.shortcut.com/scaleai/story/782703/llm-service-finetuning-api-returns-model-name-to-user-and-job-ids-are-prefixed-with-ft-instead-of-job)
[[sc-782502]](https://app.shortcut.com/scaleai/story/782502/fine-tuning-job-response-should-contain-the-fine-tuned-model)


yixu34:
Btw @felixs8696 I updated your PR description to mention the colons instead of periods (pretty sure you were copying the description from the ticket, which at the time mentioned periods) 👌🏻

docs/guides/fine_tuning.md
```diff
@@ -137,7 +152,7 @@ Once your fine-tune is finished, you will be able to start making inference requ
 from llmengine import Completion
 
 response = Completion.create(
-    model_name="ft_abc123",
+    model_name="llama-7b:airlines:2023-07-17-08-30-45",
```
yixu34, line 155: Hmm I thought we changed this to `model` @squeakymouse @yunfeng-scale @phil-scale 

In any case, not a part of this PR

yixu34 (approved):

PR #88, opened 2023-07-18
Better Model documentation

squeakymouse:


clients/python/llmengine/model.py
```diff
@@ -12,27 +12,29 @@ class Model(APIEngine):
     """
     Model API. This API is used to get, list, and delete models. Models can be created using [FineTune.create()](./#llmengine.fine_tuning.FineTune.create).
 
+    Our DTOs mention "endpoint"s, which are our concept of a physical running instance of a Model. There is a 1:1 correspondence between Models and endpoints, so you can think of them as equivalent.
```
yixu34, line 15: Maybe say "response objects" instead of "DTOs"? Granted the user is going to be technical, but probably want to be a bit more formal in this setting.

clients/python/llmengine/model.py
```diff
@@ -101,36 +119,48 @@ def list(cls) -> ListLLMEndpointsResponse:
...
+                        "inference_framework_tag": null,
+                        "num_shards": null,
+                        "quantize": null,
+                        "spec": null
+                    },
+                    {
+                        "id": null,
                         "name": "llama-7b",
-                        "model_name": "llama-7b",
+                        "model_name": null,
```
yixu34, line 135: n.b. @yunfeng-scale is in the process of removing this on the server

clients/python/llmengine/data_types.py
```diff
@@ -152,20 +156,28 @@ class GetLLMEndpointResponse(BaseModel):
...
-    id: str = Field(description="The autogenerated ID of the Launch endpoint.")
-    """The autogenerated ID of the Launch endpoint."""
-    name: str = Field(description="The name of the Launch endpoint.")
-    """The name of the Launch endpoint."""
-    model_name: str = Field(description="The name of the model.")
-    """The name of the model."""
-    source: LLMSource = Field(description="The source of the model.")
-    """The source of the model."""
+    id: Optional[str] = Field(default=None, description="(For self-hosted users) The autogenerated ID of the model.")
+    """(For self-hosted users) The autogenerated ID of the model."""
```
yixu34, line 160: nit: can we add a newline after each of these docstrings to let things visually breathe?

PR #125, opened 2023-07-18
Add back Model.create

yunfeng-scale:


yixu34 (approved):
Can you give a quick blurb as to why we're doing this? For posterity.

PR #148, opened 2023-07-20
Example notebook for fine-tuning Llama-2 7B on ScienceQA

jihan-yin:


gargutsav:
@jihan-yin Instead of the black `s3` urls for data, can you add the links to the github gists you created?

PR #160, opened 2023-07-21
File API functions

squeakymouse:
Client functions + some documentation for File API

### DIFF
--- a/clients/python/llmengine/api_engine.py
+++ b/clients/python/llmengine/api_engine.py
@@ -1,9 +1,9 @@
 # NOTICE - per Apache 2.0 license:
 # This file was copied and modified from the OpenAI Python client library: https://github.com/openai/openai-python
-from io import BufferedReader
 import json
 import os
 from functools import wraps
+from io import BufferedReader
 from typing import Any, AsyncIterable, Dict, Iterator, Optional
 
 import requests


clients/python/llmengine/data_types.py
```diff
@@ -440,3 +440,44 @@ class GetFineTuneEventsResponse(BaseModel):
...
+    id: str = Field(..., description="ID of the uploaded file.")
+    """ID of the uploaded file."""
+
+
+class GetFileResponse(BaseModel):
+    """Response object for retrieving a file."""
+
+    id: str = Field(..., description="ID of the requested file.")
+    """ID of the requested file."""
+    filename: str = Field(..., description="File name.")
```
yixu34, line 457: nit: newline spacing to let things breathe a bit?

clients/python/llmengine/file.py
```diff
@@ -0,0 +1,192 @@
...
+    GetFileContentResponse,
+    GetFileResponse,
+    ListFilesResponse,
+    UploadFileResponse,
+)
+
+
+class File(APIEngine):
+    """
+    File API. This API is used to upload private files to Scale so that fine-tunes can access them for training and validation data.
```
yixu34, line 13: We shouldn't mention Scale here, because this API will eventually work for self-hosted users too. The client can after all work in dual modes, talking to either the Scale-hosted or self-hosted LLM Engine server.

clients/python/llmengine/file.py
```diff
@@ -0,0 +1,192 @@
...
+
+    Functions are provided to upload, get, list, and delete files, as well as to get the contents of a file.
+    """
+
+    @classmethod
+    def upload(cls, file_path: str) -> UploadFileResponse:
```
yixu34, line 19: I think to start, we can do a `file` instead of a path.

clients/python/llmengine/file.py
```diff
@@ -0,0 +1,192 @@
...
+                resource_name="v1/files",
+                files=files,
+                timeout=DEFAULT_TIMEOUT,
+            )
+        return UploadFileResponse.parse_obj(response)
+
+    @classmethod
+    def get(cls, file_id: str) -> GetFileResponse:
+        """
+        Get filename and size of a file.
```
yixu34, line 58: Would suggest generalizing this to the metadata of a file, in case we add more.

clients/python/llmengine/file.py
```diff
@@ -0,0 +1,192 @@
...
+            }
+            ```
+        """
+        response = cls._get(f"v1/files/{file_id}", timeout=DEFAULT_TIMEOUT)
+        return GetFileResponse.parse_obj(response)
+
+    @classmethod
+    def list(cls) -> ListFilesResponse:
+        """
+        List all files, with information about their filenames and sizes.
```
yixu34, line 93: Same thing w.r.t metadata.

clients/python/llmengine/file.py
```diff
@@ -0,0 +1,192 @@
...
+            ```
+        """
+        response = cls._delete(
+            f"v1/files/{file_id}",
+            timeout=DEFAULT_TIMEOUT,
+        )
+        return DeleteFileResponse.parse_obj(response)
+
+    @classmethod
+    def get_content(cls, file_id: str) -> GetFileContentResponse:
```
yixu34, line 161: We might want to call this `download` to have symmetry with your `upload` function. Granted, from the _server's_ perspective, `get_content` makes more sense, but given that we're in the context of the client, I think it's ok to have the functions be client-centric.

clients/python/llmengine/file.py
```diff
@@ -0,0 +1,192 @@
...
+                "id": "file-abc123",
+                "content": "Hello world!"
+            }
+            ```
+        """
+        response = cls._get(
+            f"v1/files/{file_id}/content",
+            timeout=DEFAULT_TIMEOUT,
+        )
+        return GetFileContentResponse.parse_obj(response)
```
yixu34, line 192: Actually maybe it is fine to have this always be a string - we just need to document the expectations, e.g. if you uploaded text, then it'll ofc be that text; else if it's binary, we'll return it as a string subject to some encoding.

clients/python/llmengine/file.py
yixu34: I would say it's more important to be consistent in the client, because per above, from the server's perspective it's more about getting (extracting) content, whereas from this client, you're more specifically downloading.

yixu34: Oh wait sorry what I meant was that the client methods should be named `upload`/`download`, but it makes it easier to avoid mistakes when copying DTOs, the DTOs can follow the server. Also curious if @phil-scale or Will have thoughts on this.

yixu34 (approved):

PR #181, opened 2023-07-26
Support checkpoint_path for endpoint creation

yunfeng-scale:
```
LLM_ENGINE_BASE_PATH=xxx SCALE_API_KEY=yyy python

from llmengine import Model
Model.create(name="llama-7b-test", model="llama-7b", inference_framework_image_tag="0.9.3", checkpoint_path="s3://xxx/yyy.tar",public_inference=False)
```

### DIFF
--- a/clients/python/llmengine/data_types.py
+++ b/clients/python/llmengine/data_types.py
@@ -140,7 +140,7 @@ class CreateLLMEndpointRequest(BaseModel):
 
     quantize: Optional[Quantization] = None
     """
-    Whether to quantize the model. Only affect behavior for text-generation-inference models
+    Quantization for the LLM. Only affects behavior for text-generation-inference models
     """
 
     checkpoint_path: Optional[str] = None
--- a/clients/python/llmengine/data_types.py
+++ b/clients/python/llmengine/data_types.py
@@ -131,17 +131,27 @@ class CreateLLMEndpointRequest(BaseModel):
     # LLM specific fields
     model_name: str
     source: LLMSource = LLMSource.HUGGING_FACE
-    inference_framework: LLMInferenceFramework = LLMInferenceFramework.DEEPSPEED
+    inference_framework: LLMInferenceFramework = LLMInferenceFramework.TEXT_GENERATION_INFERENCE
     inference_framework_image_tag: str
-    num_shards: int
+    num_shards: int = 1
     """
-    Number of shards to distribute the model onto GPUs.
+    Number of shards to distribute the model onto GPUs. Only affects behavior for text-generation-inference models
+    """
+
+    quantize: Optional[Quantization] = None
+    """
+    Whether to quantize the model. Only affect behavior for text-generation-inference models
+    """
+
+    checkpoint_path: Optional[str] = None
+    """
+    Path to the checkpoint to load the model from. Only affects behavior for text-generation-inference models
     """
 
     # General endpoint fields
     metadata: Dict[str, Any]  # TODO: JSON type
     post_inference_hooks: Optional[List[str]]
-    endpoint_type: ModelEndpointType = ModelEndpointType.SYNC
+    endpoint_type: ModelEndpointType = ModelEndpointType.STREAMING
     cpus: CpuSpecificationType
     gpus: int
     memory: StorageSpecificationType
@@ -156,7 +166,10 @@ class CreateLLMEndpointRequest(BaseModel):
     high_priority: Optional[bool]
     default_callback_url: Optional[HttpUrl]
     default_callback_auth: Optional[CallbackAuth]
-    public_inference: Optional[bool] = True  # LLM endpoints are public by default.
+    public_inference: Optional[bool] = True
+    """
+    Whether the endpoint can be used for inference for all users. LLM endpoints are public by default.
+    """
 
 
 class CreateLLMEndpointResponse(BaseModel):
--- a/clients/python/llmengine/model.py
+++ b/clients/python/llmengine/model.py
@@ -12,6 +12,7 @@ from llmengine.data_types import (
     LLMSource,
     ModelEndpointType,
     PostInferenceHooks,
+    Quantization,
 )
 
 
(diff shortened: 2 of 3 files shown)


clients/python/llmengine/data_types.py
```diff
@@ -131,17 +131,27 @@ class CreateLLMEndpointRequest(BaseModel):
     # LLM specific fields
     model_name: str
     source: LLMSource = LLMSource.HUGGING_FACE
-    inference_framework: LLMInferenceFramework = LLMInferenceFramework.DEEPSPEED
+    inference_framework: LLMInferenceFramework = LLMInferenceFramework.TEXT_GENERATION_INFERENCE
```
yixu34, line 134: Is this a breaking change? Seems like we're changing default functionality. Might be fine.

clients/python/llmengine/data_types.py
```diff
@@ -131,17 +131,27 @@ class CreateLLMEndpointRequest(BaseModel):
...
-    num_shards: int
+    num_shards: int = 1
     """
-    Number of shards to distribute the model onto GPUs.
+    Number of shards to distribute the model onto GPUs. Only affects behavior for text-generation-inference models
+    """
+
+    quantize: Optional[Quantization] = None
+    """
+    Whether to quantize the model. Only affect behavior for text-generation-inference models
```
yixu34, line 143: I thought that this wasn't a boolean?

Also `Only affects`

clients/python/llmengine/data_types.py
```diff
@@ -131,17 +131,27 @@ class CreateLLMEndpointRequest(BaseModel):
...
+    checkpoint_path: Optional[str] = None
+    """
+    Path to the checkpoint to load the model from. Only affects behavior for text-generation-inference models
     """
 
     # General endpoint fields
     metadata: Dict[str, Any]  # TODO: JSON type
     post_inference_hooks: Optional[List[str]]
-    endpoint_type: ModelEndpointType = ModelEndpointType.SYNC
+    endpoint_type: ModelEndpointType = ModelEndpointType.STREAMING
```
yixu34, line 154: Does this line up with the default on the server?

yixu34 (approved):
Probably fine? Think we should get @jenkspt @sam-scale @adlam-scale to get 👀 on the API change

adlam-scale:
FYI @juliashuieh ran into trouble with the tar strategy because (I strongly suspect) of out of disk space issues. the tar file is 80G, so unzipped (the original tar file is not deleted) it takes up 160G. Then there are safetensors that are created which take up another 80G which subsequently exceeds the storage space of the pod (which is 200G). Would be nice to untar + delete or increase the hd space

clients/python/llmengine/data_types.py
yunfeng-scale: i don't think people could run build deepspeed models right now

yunfeng-scale:
> FYI @juliashuieh ran into trouble with the tar strategy because (I strongly suspect) of out of disk space issues. the tar file is 80G, so unzipped (the original tar file is not deleted) it takes up 160G. Then there are safetensors that are created which take up another 80G which subsequently exceeds the storage space of the pod (which is 200G). Would be nice to untar + delete or increase the hd space

storage size can be specified in API

clients/python/llmengine/data_types.py
yunfeng-scale: no, but this lines up with the default endpoint type for TGI

yunfeng-scale:
> FYI @juliashuieh ran into trouble with the tar strategy because (I strongly suspect) of out of disk space issues. the tar file is 80G, so unzipped (the original tar file is not deleted) it takes up 160G. Then there are safetensors that are created which take up another 80G which subsequently exceeds the storage space of the pod (which is 200G). Would be nice to untar + delete or increase the hd space

also i believe we do delete after untar, but still we'd need 160G space to untar. default storage is 96GB

clients/python/llmengine/model.py
```diff
@@ -68,6 +72,13 @@ def create(
             num_shards (`int`):
                 Number of shards for the LLM. When bigger than 1, LLM will be sharded
                 to multiple GPUs. Number of GPUs must be larger than num_shards.
+                Only affects behavior for text-generation-inference models
+
+            quantize (`Optional[Quantization]`):
+                Quantization for the LLM. Only affects behavior for text-generation-inference models
+
+            checkpoint_path (`Optional[str]`):
```
jenkspt, line 80: Can I get a bit more info here (or link to info) on the requirements for the checkpoint? i.e does it need to be a .tar file? do weights need to be saved as safetensors?

clients/python/llmengine/model.py
```diff
@@ -28,12 +29,15 @@ class Model(APIEngine):
     @assert_self_hosted
     def create(
```
jenkspt, line 30: How do I specify the instance type? or are all models on the same instance (I need at least g5.12xl)

jenkspt, line 30: `name` and `model` parameters need to be fixed in the docstring

jenkspt:
I get this is out-of-scope for this PR - but not sure where to put it. Can remove the requirement to tar our checkpoints?

jenkspt:
Seems I can't use the default `min_workers=0`

```
BadRequestError                           Traceback (most recent call last)
Cell In[4], line 1
----> 1 out = Model.create(name='test-falcon-7b-deploy', model='falcon-7b', checkpoint_path='s3://scale-ml/users/penn/deploy/test-falcon-7b-deploy.tar', inference_framework_image_tag="0.9.3")

File ~/llm-engine/clients/python/llmengine/api_engine.py:27, in assert_self_hosted.<locals>.inner(*args, **kwargs)
     25 if SPELLBOOK_API_URL == LLM_ENGINE_BASE_PATH:
     26     raise ValueError("This feature is only available for self-hosted users.")
---> 27 return func(*args, **kwargs)

File ~/llm-engine/clients/python/llmengine/model.py:195, in Model.create(cls, name, model, inference_framework_image_tag, source, inference_framework, num_shards, quantize, checkpoint_path, cpus, memory, storage, gpus, min_workers, max_workers, per_worker, endpoint_type, gpu_type, high_priority, post_inference_hooks, default_callback_url, public_inference, labels)
    168             post_inference_hooks_strs.append(hook)
    170 request = CreateLLMEndpointRequest(
    171     name=name,
    172     model_name=model,
   (...)
    193     public_inference=public_inference,
    194 )
--> 195 response = cls.post_sync(
    196     resource_name="v1/llm/model-endpoints",
    197     data=request.dict(),
    198     timeout=DEFAULT_TIMEOUT,
    199 )
    200 return CreateLLMEndpointResponse.parse_obj(response)

File ~/llm-engine/clients/python/llmengine/api_engine.py:96, in APIEngine.post_sync(cls, resource_name, data, timeout)
     88 response = requests.post(
     89     os.path.join(LLM_ENGINE_BASE_PATH, resource_name),
     90     json=data,
   (...)
     93     auth=(api_key, ""),
     94 )
     95 if response.status_code != 200:
---> 96     raise parse_error(response.status_code, response.content)
     97 payload = response.json()
     98 return payload

BadRequestError: Requested min workers 0 too low
```

using `min_workers=1` works fine

PR #185, opened 2023-07-29
Add llama 2 70B in model zoo

yunfeng-scale:


acmatscale:
Also need to update docs

PR #196, opened 2023-08-04
adding download api to launch client, updating example

ian-scale:


yunfeng-scale:
can you share screenshot of newly added page?

clients/python/llmengine/model.py
```diff
@@ -366,3 +368,50 @@ def delete(cls, model: str) -> DeleteLLMEndpointResponse:
...
+        The user can then download these urls to obtain the fine-tuned model. If called 
+        on a nonexistent model, an error will be thrown.
+
+        Args:
+            model_name (`str`):
+                name of the fine-tuned model 
+            download_format (`str`): 
+                download format requested (default=huggingface) 
+        Returns:
+            DownloadModelResponse: an object that contains a dictionary of filenames, urls from which to download the model weights
```
yunfeng-scale, line 393: can you mention those urls are presigned s3 urls and what's the expiration?

ian-scale: yes I'll make that explicit!

ian-scale:
> can you share screenshot of newly added page?

i.e. the .ipynb example?

clients/python/llmengine/model.py
```diff
@@ -366,3 +368,51 @@ def delete(cls, model: str) -> DeleteLLMEndpointResponse:
         """
         response = cls._delete(f"v1/llm/model-endpoints/{model}", timeout=DEFAULT_TIMEOUT)
         return DeleteLLMEndpointResponse.parse_obj(response)
+
+    @classmethod
+    def download(
+        cls,
+        model_name: str,
+        download_format: str = "huggingface",
```
yixu34, line 376: It seems like we've used `hugging_face` elsewhere in LLM Engine. I think the motivation for this was due to Hugging Face being two words?
image

clients/python/llmengine/model.py
```diff
@@ -366,3 +368,51 @@ def delete(cls, model: str) -> DeleteLLMEndpointResponse:
...
+        on a nonexistent model, an error will be thrown.
+
+        Args:
+            model_name (`str`):
+                name of the fine-tuned model
+            download_format (`str`):
+                download format requested (default=huggingface)
+        Returns:
+            DownloadModelResponse: an object that contains a dictionary of filenames, urls from which to download the model weights.
+            The urls are presigned urls that grant temporary access and expire after an hour.
```
yixu34, line 394: Think this is OK for now, but not sure if this is technically true in all contexts, e.g. self-hosting.

yixu34 (commented):
Normally I'd suggest adding a helper function like `Model.download` instead of relying on these code snippets, but given that we don't actually want to make this _too_ easy, I think that's fine.

clients/python/llmengine/model.py
ian-scale: This make sense, I'm actually not exactly sure what the behavior would be in a self-hosted context. We can think about this more as it becomes relevant in the future?

clients/python/llmengine/model.py
ian-scale: oh okay, that's a relatively easy fix (currently we're not actually doing anything with the format token anyway)

PR #200, opened 2023-08-07
Add documentation on pointing llmengine client to self-hosted infrastructure

ruizehung-scale:
Add documentation on pointing `llmengine` client to self-hosted infrastructure.

### DIFF
--- a/docs/guides/self_hosting.md
+++ b/docs/guides/self_hosting.md
@@ -205,7 +205,7 @@ You should get a response similar to:
 ### Pointing LLM Engine client to use self-hosted infrastructure
 The `llmengine` client makes requests to Scale AI's hosted infrastructure by default. You can have `llmengine` client make requests to your own self-hosted infrastructure by setting the `LLM_ENGINE_BASE_PATH` environment variable to the URL of the `llm-engine` pod. 
 
-The exact URL of `llm-engine` pod depends on your Kubernetes cluster networking setup. The domain is specified in `config.values.infra.dns_host_domain` in the helm chart values config file. Using `charts/llm-engine/values_sample.yaml` as an example, you would 
+The exact URL of `llm-engine` pod depends on your Kubernetes cluster networking setup. The domain is specified at `config.values.infra.dns_host_domain` in the helm chart values config file. Using `charts/llm-engine/values_sample.yaml` as an example, you would 
 ```bash
 export LLM_ENGINE_BASE_PATH=https://domain.llm-engine.com
 ```
--- a/docs/guides/self_hosting.md
+++ b/docs/guides/self_hosting.md
@@ -207,5 +207,5 @@ The `llmengine` client makes requests to Scale AI's hosted infrastructure by def
 
 The exact URL of `llm-engine` service depends on your Kubernetes cluster networking setup. The domain is specified at `config.values.infra.dns_host_domain` in the helm chart values config file. Using `charts/llm-engine/values_sample.yaml` as an example, you would do:
 ```bash
-export LLM_ENGINE_BASE_PATH=https://domain.llm-engine.com
+export LLM_ENGINE_BASE_PATH=https://llm-engine.domain.com
 ```
--- a/docs/guides/self_hosting.md
+++ b/docs/guides/self_hosting.md
@@ -203,4 +203,4 @@ You should get a response similar to:
 ```
 
 ### Pointing LLM Engine client to use self-hosted infrastructure
-The `llmengine` client makes requests to Scale AI's hosted infrastructure by default. You can have `llmengine` client make requests to your own self-hosted infrastructure by setting the `LLM_ENGINE_BASE_PATH` environment variable to the url of the `llm-engine` pod. The exact url of `llm-engine` pod depends on your Kubernetes cluster networking setup.
+The `llmengine` client makes requests to Scale AI's hosted infrastructure by default. You can have `llmengine` client make requests to your own self-hosted infrastructure by setting the `LLM_ENGINE_BASE_PATH` environment variable to the URL of the `llm-engine` pod. The exact URL of `llm-engine` pod depends on your Kubernetes cluster networking setup.
--- a/docs/guides/self_hosting.md
+++ b/docs/guides/self_hosting.md
@@ -201,3 +201,6 @@ You should get a response similar to:
 ```
 {"status":"SUCCESS","outputs":[{"text":". Tell me a joke about AI. Tell me a joke about AI. Tell me a joke about AI. Tell me","num_completion_tokens":30}],"traceback":null}
 ```
+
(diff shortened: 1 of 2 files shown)


docs/guides/self_hosting.md
```diff
@@ -200,4 +200,7 @@ $ curl -X POST 'http://localhost:5000/v1/llm/completions-sync?model_endpoint_nam
 You should get a response similar to:
 ```
 {"status":"SUCCESS","outputs":[{"text":". Tell me a joke about AI. Tell me a joke about AI. Tell me a joke about AI. Tell me","num_completion_tokens":30}],"traceback":null}
-```
\ No newline at end of file
+```
+
+### Pointing LLM Engine client to use self-hosted infrastructure
+The `llmengine` client makes requests to Scale AI's hosted infrastructure by default. You can have `llmengine` client make requests to your own self-hosted infrastructure by setting the `LLM_ENGINE_BASE_PATH` environment variable to the URL of the `llm-engine` pod. The exact URL of `llm-engine` pod depends on your Kubernetes cluster networking setup.
```
ruizehung-scale, line 206: I'm not 100% sure if this would just work out of the box as Spellbook API might be different from the API exposed by llm-engine gateway?

ruizehung-scale: Does having people use https://github.com/scaleapi/launch-python-client and point the `gateway_endpoint` to self-hosted llm-engine be better?

seanshi-scale: I'd have thought that we'd have wanted to have `llmengine` client be able to point to `llmengine` server directly? in the sense that if this doesn't work OOB we should make it work OOB?

IMO having people use `launch-python-client` seems kinda ugly

yixu34: Let's not mention `launch-python-client` here. FWIW, the EGP APIs do have a `/llm`-prefixed set of APIs that are at parity with the LLM Engine APIs, so having the LLM Engine client point to either the self-hosted or EGP-hosted one should still work. There is a _separate_ set of EGP-specific Completion APIs, which is not at play here. cc @felixs8696

ruizehung-scale: Ok! Then I'll keep the documentation as it is and only use `llmengine` client

yunfeng-scale: > Spellbook API might be different from the API exposed by llm-engine gateway

we intentionally keep them the same.

yunfeng-scale (approved):

docs/guides/self_hosting.md
yixu34, line 206: > The exact URL of `llm-engine` pod depends on your Kubernetes cluster networking setup.

This might be true, but I think we have some default in our helm chart? @song-william @phil-scale 

Also, let's add a code snippet example, which assumes this default.

ruizehung-scale: @yixu34 yep you're right! The default cluster domain is specified here
https://github.com/scaleapi/llm-engine/blob/point-client-to-self-hosted/charts/llm-engine/values_sample.yaml#L99

docs/guides/self_hosting.md
```diff
@@ -200,4 +200,12 @@ $ curl -X POST 'http://localhost:5000/v1/llm/completions-sync?model_endpoint_nam
 You should get a response similar to:
 ```
 {"status":"SUCCESS","outputs":[{"text":". Tell me a joke about AI. Tell me a joke about AI. Tell me a joke about AI. Tell me","num_completion_tokens":30}],"traceback":null}
+```
+
+### Pointing LLM Engine client to use self-hosted infrastructure
+The `llmengine` client makes requests to Scale AI's hosted infrastructure by default. You can have `llmengine` client make requests to your own self-hosted infrastructure by setting the `LLM_ENGINE_BASE_PATH` environment variable to the URL of the `llm-engine` pod.
```
yixu34, line 206: Would suggest replacing "pod" with "service". pod is too specific and low-level - in fact, there may be many pods for a given service.

docs/guides/self_hosting.md
```diff
@@ -200,4 +200,12 @@ $ curl -X POST 'http://localhost:5000/v1/llm/completions-sync?model_endpoint_nam
 You should get a response similar to:
 ```
 {"status":"SUCCESS","outputs":[{"text":". Tell me a joke about AI. Tell me a joke about AI. Tell me a joke about AI. Tell me","num_completion_tokens":30}],"traceback":null}
+```
+
+### Pointing LLM Engine client to use self-hosted infrastructure
+The `llmengine` client makes requests to Scale AI's hosted infrastructure by default. You can have `llmengine` client make requests to your own self-hosted infrastructure by setting the `LLM_ENGINE_BASE_PATH` environment variable to the URL of the `llm-engine` pod. 
+
+The exact URL of `llm-engine` pod depends on your Kubernetes cluster networking setup. The domain is specified at `config.values.infra.dns_host_domain` in the helm chart values config file. Using `charts/llm-engine/values_sample.yaml` as an example, you would
```
yixu34, line 208: Same here, pod -> service.

yixu34: Also finish this sentence, e.g. `you would do:`

docs/guides/self_hosting.md
```diff
@@ -200,4 +200,12 @@ $ curl -X POST 'http://localhost:5000/v1/llm/completions-sync?model_endpoint_nam
...
 ```
 {"status":"SUCCESS","outputs":[{"text":". Tell me a joke about AI. Tell me a joke about AI. Tell me a joke about AI. Tell me","num_completion_tokens":30}],"traceback":null}
+```
+
+### Pointing LLM Engine client to use self-hosted infrastructure
+The `llmengine` client makes requests to Scale AI's hosted infrastructure by default. You can have `llmengine` client make requests to your own self-hosted infrastructure by setting the `LLM_ENGINE_BASE_PATH` environment variable to the URL of the `llm-engine` service. 
+
+The exact URL of `llm-engine` service depends on your Kubernetes cluster networking setup. The domain is specified at `config.values.infra.dns_host_domain` in the helm chart values config file. Using `charts/llm-engine/values_sample.yaml` as an example, you would do:
+```bash
+export LLM_ENGINE_BASE_PATH=https://domain.llm-engine.com
```
seanshi-scale, line 210: nit: should this be `https://llm-engine.domain.com`? (in addition to changing the value inside `values_sample.yaml`) 

since it feels like `llm-engine` should be some subdomain of `domain.com`, not the other way around, if users are self-hosting at `domain.com`.

ruizehung-scale: Make sense! Updated!

seanshi-scale (approved):

yixu34 (approved):

PR #202, opened 2023-08-07
adding status field to model get response

ian-scale:
Per request to add status of model endpoint to `Model.get()`.

### DIFF
--- a/clients/python/llmengine/data_types.py
+++ b/clients/python/llmengine/data_types.py
@@ -202,7 +202,7 @@ class GetLLMEndpointResponse(BaseModel):
     """The source of the model, e.g. Hugging Face."""
 
     status: ModelEndpointStatus = Field(description="The status of the model.")
-    """The status of the model."""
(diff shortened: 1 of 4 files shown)


yunfeng-scale (approved):

yunfeng-scale:
with this, can you update docs like `getting_started.md` that how to wait for model deployment to finish?

clients/python/llmengine/data_types.py
```diff
@@ -201,6 +201,9 @@ class GetLLMEndpointResponse(BaseModel):
     source: LLMSource = Field(description="The source of the model, e.g. Hugging Face.")
     """The source of the model, e.g. Hugging Face."""
 
+    status: ModelEndpointStatus = Field(description="The status of the model.")
+    """The status of the model."""
```
yixu34, line 205: Would be good to list the possible values.

yixu34 (approved):

clients/python/llmengine/data_types.py
ian-scale: done!

PR #211, opened 2023-08-16
Faster s5cmd download

yunfeng-scale:
- don't depend on private docker image
- download from tar file or folder
- optimize s5cmd parameters


yixu34 (approved):

yunfeng-scale:
nope don't want to do this PR on a fork

PR #223, opened 2023-08-25
Fix docs building

phil-scale:


yunfeng-scale (approved):

docs/api/data_types.md
```diff
@@ -16,16 +16,37 @@
             - num_completion_tokens
 
 ::: llmengine.CompletionSyncResponse
+    selection:
```
yunfeng-scale, line 19: hmm, do we must list these one by one?

phil-scale: There's a configuration in `mkdocs.yml` that allows members to be listed automatically, but the trade off is that we have no control over the order of the members.

Issue #209, opened 2023-08-15
Bug: API calls don't work on Windows due to os.path.join


f-linus:
Hey everyone. This stems from an older, closed issue: https://github.com/scaleapi/llm-engine/issues/170.

The post_sync method in api_engine.py concatenates parts of the URL to make the API call against using the os.path.join Python method.

See:
```python
    @classmethod
    def post_sync(cls, resource_name: str, data: Dict[str, Any], timeout: int) -> Dict[str, Any]:
        api_key = get_api_key()
        response = requests.post(
            os.path.join(LLM_ENGINE_BASE_PATH, resource_name),
            json=data,
            timeout=timeout,
            headers={"x-api-key": api_key},
            auth=(api_key, ""),
        )
        if response.status_code != 200:
            raise parse_error(response.status_code, response.content)
        payload = response.json()
        return payload
```

However, that method is designed to join paths on file systems depending on the OS Python is being executed on. When on Windows, this results in the path being joined into a URL where one slash is actually a backslash since that is the "divider" used on the Windows file system.

Long story short: bug that breaks all API calls when using package on Windows since os.path.join is used which puts a backslash instead of a slash into the endpoint URL

saiatmakuri:
#225 should fix the issue - I've tested on a Windows machine and it worked for me!

Issue #170, opened 2023-07-25
FineTune.create - NotFoundError (API endpoint seems to through 404)


f-linus:
When calling FineTune.create with an arbitrary model and CSV a "NotFoundError" is thrown. When investigating, the Scale AI Spellbook API endpoint at https://api.spellbook.scale.com/v1/llm/fine-tune throws an 404.

Used version: 0.0.0b5

Code:
```Pyhon
from llmengine import FineTune
import os
import yaml

if __name__ == "__main__":
    with open("environments/spellbook_api_key.yaml", "r") as stream:
        os.environ.update(yaml.safe_load(stream))

    response = FineTune.create(
        model="llama-2-7b",
        training_file="storage.googleapis.com/llama2-finetune-test/finetune_dataset.csv",
    )

    print(response.json())
```

Error:
![image](https://github.com/scaleapi/llm-engine/assets/84272959/1d25931e-31b7-4f4f-bad0-0047fe016e51)

KensingtonOscupant:
Seeing this too  :/

yixu34:
Looking into it

f-linus:
Error still persists after upgrading to 0.0.0.b7

yixu34:
I'm going to close this out for now - @f-linus feel free to open if you're seeing the issue persist, after accounting for the backslash in the path.

PR #225, opened 2023-08-25
Fix API calls on Windows

saiatmakuri:
`os.path.join` is system-dependent, so it adds backslashes on Windows; this creates invalid routes. 

instead use `urllib.parse.urljoin` that is system independent and is built for joining URLs.

https://github.com/scaleapi/llm-engine/issues/170
https://github.com/scaleapi/llm-engine/issues/209

### DIFF
--- a/clients/python/llmengine/__init__.py
+++ b/clients/python/llmengine/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.0.beta11"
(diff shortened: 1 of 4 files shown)


yunfeng-scale (approved):

song-william (approved):
Were you able to test this on a windows machine?

saiatmakuri:
> Were you able to test this on a windows machine?

yup. got a Windows EC2 instance

yixu34 (approved):
Remember to publish a new client on pypi!

PR #227, opened 2023-08-25
Add actual pytests to integration tests

phil-scale:


integration_tests/rest_api_utils.py
phil-scale, line 1: mostly just copied from `e2e_test_v1.py` - probably should clean this up in a future PR

charts/model-engine/values_circleci.yaml
```diff
@@ -136,7 +136,7 @@ config:
 
       billing_queue_arn: none
       cache_redis_url: redis://redis-message-broker-master.default/15
-      s3_file_llm_fine_tune_repository: "s3://$CIRCLECI_AWS_S3_BUCKET"
+      s3_file_llm_fine_tune_repository: "s3://model-engine-integration-tests/fine_tune_repository/circleci"
```
yixu34, line 139: If we're ditching this test anyway, can we remove this? Don't think we want to have a public bucket just to support this test.

integration_tests/test_batch_jobs.py
```diff
@@ -0,0 +1,25 @@
...
+    USER_ID_0,
+    cancel_batch_job,
+    create_batch_job,
+    create_docker_image_batch_job,
+    get_or_create_docker_image_batch_job_bundle,
+)
+from .test_bundles import model_bundles  # noqa
+
+
+def test_di_batch_jobs(model_bundles) -> None:  # noqa
```
yixu34, line 14: If we're skipping this, should we do `pytest.mark.skip`? (does this run via `pytest`?)

integration_tests/test_docs.py
```diff
@@ -0,0 +1,210 @@
...
+
+@pytest.fixture()
+def seed() -> int:
+    """Returns a random seed between 0 and 999, inclusive."""
+    return random.randint(0, 999)
+
+
+@pytest.fixture()
+def integration_test_user_id() -> str:
+    return "62bc820451dbea002b1c5421"
```
yixu34, line 61: Should this be here?

integration_tests/test_file.py
```diff
@@ -0,0 +1,27 @@
+from .rest_api_utils import (  # list_files, delete_file_by_id,
+    get_file_by_id,
+    get_file_content_by_id,
+    upload_file,
+)
+
+
+def test_files() -> None:
+    user = "62bc820451dbea002b1c5421"  # CDS needs proper user ID
```
yixu34, line 9: Not sure if this should be there.

model-engine/model_engine_server/infra/gateways/resources/templates/service_template_config_map_circleci.yaml
```diff
@@ -2807,11 +2812,13 @@ data:
...
                 - name: LAUNCH_SERVICE_TEMPLATE_FOLDER
                   value: /workspace/model-engine/model_engine_server/infra/gateways/resources/templates
+                - name: CIRCLECI
+                  value: "true"
                 - name: DD_VERSION
                   value: ${GIT_TAG}
                 - name: GIT_TAG
                   value: ${GIT_TAG}
-              imagePullPolicy: IfNotPresent
+              imagePullPolicy: Always
```
yixu34, line 2821: Hm why do we always want to pull?

yixu34 (approved):

integration_tests/test_docs.py
louisbarrett: We should switch to some unit test specific ID here...

PR #228, opened 2023-08-26
Add vLLM as an inference framework

yunfeng-scale:
Add a dockerfile for vLLM, also API change to create such an endpoint


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -401,6 +425,86 @@ async def create_deepspeed_bundle(
...
+        num_shards: int,
+        checkpoint_path: Optional[str],
+    ):
+        command = []
+
+        max_num_batched_tokens = 2560  # vLLM's default
+        if "llama-2" in model_name:
+            max_num_batched_tokens = 4096  # Need to be bigger than model's context window
+
+        assert checkpoint_path is not None, "Checkpoint path must be provided for vLLM"
```
yixu34, line 443: Should this be `Optional` if we're really requiring it?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -401,6 +425,86 @@ async def create_deepspeed_bundle(
...
+            final_weights_folder = "model_files"
+            subcommands = []
+
+            s5cmd = "./s5cmd"
+
+            if base_path.endswith(".tar"):
+                # If the checkpoint file is a tar file, extract it into final_weights_folder
+                subcommands.extend(
+                    [
+                        f"{s5cmd} cp {checkpoint_path} .",
```
yixu34, line 456: Probably fine for now, but if we have 3 frameworks, we might want to refactor this s5cmd piece into a common component.

model-engine/model_engine_server/inference/vllm/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM nvcr.io/nvidia/pytorch:22.12-py3
+
+RUN pip uninstall torch -y
+RUN pip install vllm==0.1.3
+RUN pip install ray[air]==2.6.3
```
yixu34, line 5: nit: combine these `pip install`s into a single command?

model-engine/model_engine_server/inference/vllm/build_and_upload_image.sh
```diff
@@ -0,0 +1,15 @@
...
+
+set -eo pipefail
+
+if [ -z "$1" ]; then
+  echo "Must supply the image tag"
+  exit 1;
+fi
+
+IMAGE_TAG=$1
+aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin 692474966980.dkr.ecr.us-west-2.amazonaws.com
```
yixu34, line 13: Internal stuff?

Think we need to notify our sponsor cc @louisbarrett 🥃 🥴

model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -0,0 +1,100 @@
...
+    final_output = None
+    async for request_output in results_generator:
+        if await request.is_disconnected():
+            # Abort the request if the client disconnects.
+            await engine.abort(request_id)
+            return Response(status_code=499)
+        final_output = request_output
+
+    assert final_output is not None
+    print(final_output)
```
yixu34, line 73: Do we need to keep this?

model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -0,0 +1,100 @@
+import argparse
+import json
+from typing import AsyncGenerator
+
+import uvicorn
+from fastapi import BackgroundTasks, FastAPI, Request
+from fastapi.responses import Response, StreamingResponse
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.engine.async_llm_engine import AsyncLLMEngine
```
yixu34, line 9: lol

model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -0,0 +1,100 @@
...
+from vllm.sampling_params import SamplingParams
+from vllm.utils import random_uuid
+
+TIMEOUT_KEEP_ALIVE = 5  # seconds.
```
yixu34, line 13: Do we need concurrency limiting too?

yixu34 (changes requested):
Internal stuff?

model-engine/model_engine_server/inference/vllm/build_and_upload_image.sh
yunfeng-scale: oh looks like i made change in wrong place

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: sure

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: hmm it might be convenient for external users to make it actually optional

model-engine/model_engine_server/inference/vllm/vllm_server.py
yunfeng-scale: concurrency limiting lives in http forwarder

model-engine/model_engine_server/inference/vllm/vllm_server.py
yunfeng-scale: nope

model-engine/model_engine_server/inference/vllm/build_and_upload_image.sh
```diff
@@ -1,15 +1,21 @@
...
   exit 1;
 fi
 
-IMAGE_TAG=$1
-aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin 692474966980.dkr.ecr.us-west-2.amazonaws.com
-DOCKER_BUILDKIT=1 docker build -t 692474966980.dkr.ecr.us-west-2.amazonaws.com/vllm:$IMAGE_TAG .
-docker push 692474966980.dkr.ecr.us-west-2.amazonaws.com/vllm:$IMAGE_TAG
+IMAGE_TAG=$2
+ACCOUNT=$1
+aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin $ACCOUNT.dkr.ecr.us-west-2.amazonaws.com
```
yixu34, line 19: Might want to pass in a full ECR URL instead of just the account, because this leaks info about the region. But think this is ok for now.

yixu34 (approved):

PR #229, opened 2023-08-28
Sync scale from zero, part 1

seanshi-scale:
Emit autoscaling metrics from gateway
Have sync endpoints poll + wait if it gets upstream not found
Next pr: make endpoints created with the keda object if we're scaling from zero
Also change how we're doing retries for sync endpoints in general, and add some optional fields to have some user control over retry behavior.

Note: this scales pods from 0 to 1, not 1 to N. Autoscaling works by looking at the length of a redis list, and the implementation intentionally keeps the list at length 1. We also set a timeout for the redis list at 1 minute.


yunfeng-scale (approved):

yunfeng-scale:
unit test failed?

PR #230, opened 2023-08-28
Sync scale from zero part 2

seanshi-scale:
-Test create, get, delete
-Create and get and delete work
-Autoscaling does seem to work

-also add in some code for allowing artifactlike endpoints to be able to build

-todo docker image fine tuning entrypoint probably needs high-priority turned on


model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -196,6 +196,7 @@ def load_k8s_yaml(key: str, substitution_kwargs: ResourceArguments) -> Dict[str,
     yaml_str = Template(template_str).substitute(**substitution_kwargs)
     try:
         yaml_obj = yaml.safe_load(yaml_str)
+        print(yaml_obj)
```
seanshi-scale, line 199: todo rm

model-engine/model_engine_server/infra/repositories/ecr_docker_repository.py
```diff
@@ -25,6 +25,10 @@ def build_image(self, image_params: BuildImageRequest) -> BuildImageResponse:
         if image_params.requirements_folder:
             folders_to_include.append(image_params.requirements_folder)
 
+        dockerfile_root_folder = image_params.dockerfile.split("/")[0]
```
seanshi-scale, line 28: todo split out into own pr

charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -396,6 +396,36 @@ data:
             target:
               type: Value
               averageValue: ${CONCURRENCY}
+  # TODO maybe fix keda scaled object
```
seanshi-scale, line 399: TODO this seems to work

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -599,6 +600,45 @@ async def _create_vpa(vpa: Dict[str, Any], name: str) -> None:
                 logger.exception("Got an exception when trying to apply the VerticalPodAutoscaler")
                 raise
 
+    @staticmethod
+    async def _create_keda_scaled_object(scaled_object: Dict[str, Any], name: str) -> None:
+        # TODO test
```
seanshi-scale, line 605: seems to work e2e

seanshi-scale: at least create works, update idk

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -987,14 +1027,36 @@ async def _delete_hpa(endpoint_id: str, deployment_name: str) -> bool:
                         logger.exception(
                             f"Deletion of HorizontalPodAutoscaler {k8s_resource_group_name} failed"
                         )
-                        return False
+                    return False
```
seanshi-scale, line 1030: this does change the semantics of `_delete_hpa()`

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1102,19 +1164,36 @@ async def _create_or_update_resources(
...
-                sqs_queue_url=sqs_queue_url_str,
-                endpoint_resource_name="horizontal-pod-autoscaler",
-                api_version=api_version,
-            )
-            hpa_template = load_k8s_yaml("horizontal-pod-autoscaler.yaml", hpa_arguments)
-            await self._create_hpa(
-                hpa=hpa_template,
-                name=k8s_resource_group_name,
-            )
+            # TODO select between hpa and keda scaled object somehow
```
seanshi-scale, line 1167: todo rm

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: todo maybe just try both deletions then OR the function results?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -599,6 +599,44 @@ async def _create_vpa(vpa: Dict[str, Any], name: str) -> None:
...
+                # one. See _create_vpa for more details.
+                existing_scaled_object = await custom_objects_api.get_namespaced_custom_object(
+                    group="keda.sh",
+                    version="v1alpha1",
+                    namespace=hmi_config.endpoint_namespace,
+                    plural="scaledobjects",
+                    name=name,
+                )
+                new_scaled_object = deep_update(existing_scaled_object, scaled_object)
+                await custom_objects_api.replace_namespaced_custom_object(
```
yixu34, line 628: Should we do replace? Or patch? I just remember that replacing caused rolling restarts to not work, which doesn't apply here, but if it's all the same, maybe just patch?

yixu34 (approved):

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: I saw that other resources using the custom_objects_api apparently needed to use replace, so am doing that here as well. Separately there's an option `restoreToOriginalReplicaCount ` [here](https://keda.sh/docs/2.11/concepts/scaling-deployments/#advanced) that is default false, and which I haven't set, so I think what should happen is that as the ScaledObject gets deleted the deployment doesn't get affected, so we shouldn't get downtime here at least.

model-engine/model_engine_server/infra/gateways/resources/templates/service_template_config_map_circleci.yaml
```diff
@@ -2558,6 +2558,47 @@ data:
             target:
               type: Value
               averageValue: ${CONCURRENCY}
+  keda-scaled-object.yaml: |-
```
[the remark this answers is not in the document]
seanshi-scale: tbh not sure, maybe it's used for the integration tests in circleci?
in any case if we're not creating keda scaled objects in minikube I think we should be fine and won't break tests, but once we do we might have to helm install keda into minikube or something (or support some more options so keda doesn't become a hard dependency for self hosting)

PR #231, opened 2023-08-29
add file upload guidance to fine-tune docs

saiatmakuri:
fine-tune guide specifies example of using llm engine's file upload

### DIFF
--- a/docs/guides/fine_tuning.md
+++ b/docs/guides/fine_tuning.md
@@ -105,7 +105,7 @@ with open('customer_service_data.csv', 'w', newline='') as file:
 
 Currently, data needs to be uploaded to either a publicly accessible web URL or to LLM Engine's private file server so that it can be read for fine-tuning. Publicly accessible HTTP and HTTPS URLs are currently supported.
 
-To privately share data with the LLM Engine API, use LLM Engine's [File.upload](../../api/python_client/#llmengine.File.upload) API. You can upload data in local file to LLM Engine's private file server and then use the returned file ID to reference your data in the FineTune API.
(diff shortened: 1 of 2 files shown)


docs/guides/fine_tuning.md
```diff
@@ -137,8 +141,7 @@ from llmengine import FineTune
 
 response = FineTune.create(
     model="llama-2-7b",
-    training_file="s3://my-bucket/path/to/training-file.csv",
-    validation_file="s3://my-bucket/path/to/validation-file.csv",
+    training_file="file-7DLVeLdN2Ty4M2m",
```
yixu34, line 144: Can we keep a `validation_file` in our example?

docs/guides/fine_tuning.md
```diff
@@ -103,14 +103,18 @@ with open('customer_service_data.csv', 'w', newline='') as file:
...
+
+To privately share data with the LLM Engine API, use LLM Engine's [File.upload](../../api/python_client/#llmengine.File.upload) API. You can upload data in local file to LLM Engine's private file server and then use the returned file ID to reference your data in the FineTune API.
+
+=== "Upload to LLM Engine's private file server"
+
+```python
+from llmengine import File
+
+response = File.upload(open("customer_service_data.csv", "r"))
+print(response.json())
```
yixu34, line 116: Perhaps we can add a comment that says it prints out a particular ID (e.g. you write `file-7DLVeLdN2Ty4M2m` below)? Then we can just reference that ID below. Well, IDs plural, if we do this for the `validation_file` as well.

saiatmakuri: added note above

yixu34 (approved):

PR #234, opened 2023-08-29
Ianmacleod/completion sync error throws 4xx

ian-scale:
fixing errors so that we don't run into so many 5xx errors when TGI returns an issue due to input prompt window being too large. 

tested with `just up` locally to make sure behavior is consistent when calling with a prompt that exceeds the max number of tokens specified by TGI tokenizer.


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -636,7 +636,10 @@ def model_output_to_completion_output(
                 )
             except Exception as e:
                 logger.exception(f"Error parsing text-generation-inference output {model_output}")
-                raise e
+                if 'generated_text' not in model_output:
+                    raise ObjectHasInvalidValueException(
```
yixu34, line 640: Hmm yeah I'm not sure it necessarily follows that a user passed a bad input, if we observe `if 'generated_text' not in model_output:`. It seems like we're doing this empirically instead of going against TGI's API (not sure if they document this behavior). Perhaps the original exception that gets thrown can tell us something? @yunfeng-scale thoughts?

yunfeng-scale: i don't think TGI's error behavior is documented, they just send back a string when error

can you change error handling for streaming as well?

ian-scale: yeah I'll update the error handling for streaming as well

ian-scale: The initial error message from TGI is a stringified json message: "{'error': 'Input validation error: `inputs` must have less than 2048 tokens. Given: 2687', 'error_type': 'validation'}". I could parse this error and return it to the user wrapped as an InputValidationError? Does that sound better? @yixu34 @yunfeng-scale

yixu34: Generally not ideal to try and parse strings, and it's instead preferable to rely on structured fields. But in the absence of that, think this is fine for now. It's at least more precise than inferring semantics like what's currently being done in this PR?

I'd be in favor of doing the parse, putting a TODO, and revisiting this.

yunfeng-scale: i think we could try to parse it as an error and return more precise content. the current error `Error parsing text-generation-inference output` does not really capture what the error is and might confuse user

ian-scale: Okay added this logic for completion sync

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -919,15 +922,21 @@ async def execute(
...
+                                text=result["result"]["token"]["text"],
+                                finished=finished,
+                                num_completion_tokens=num_completion_tokens,
+                                token=token,
+                            ),
+                        )
+                    except Exception as e:  # if result["result"]["token"]["text"] doesn't exist 
+                        logger.exception(f"Error parsing text-generation-inference stream output. Error message: {e}")
+                        raise ObjectHasInvalidValueException(
+                            f"Error parsing text-generation-inference stream output."
```
ian-scale, line 938: @yunfeng-scale the error that's returned to the user here if I were to parse the error is just a KeyError, which is confusing to return (stems from the fact that result["result"]["token"]["text"] doesn't actually exist), do you think handling it this way is fine?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -635,8 +635,11 @@ def model_output_to_completion_output(
                     tokens=tokens,
                 )
             except Exception as e:
-                logger.exception(f"Error parsing text-generation-inference output {model_output}")
-                raise e
+                logger.exception(f"Error parsing text-generation-inference output {model_output}. Error message: {json.loads(e)['error']}")
+                if 'generated_text' not in model_output:
+                    raise ObjectHasInvalidValueException(
+                        f"Error parsing text-generation-inference output {model_output}. Error message: {json.loads(e)['error']}"
```
yunfeng-scale, line 641: my suggestion is:
```
if model_output.get("error_type") == "validation":
  raise InvalidRequestException(model_output.get("error") # trigger a 400
else:
  raise Exception(model_output.get("error")) # some exception that will trigger a 500 so user can retry
```

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: what about just check whether result contains "error" as a key

model-engine/model_engine_server/core/domain_exceptions.py
```diff
@@ -57,3 +57,8 @@ class ReadOnlyDatabaseException(DomainException):
     """
     Thrown if the server attempted to write to a read-only database.
     """
+
+class InvalidRequestException(DomainException):
```
yunfeng-scale, line 61: it already exists in model_engine_server.domain.exceptions

yunfeng-scale: not sure difference between `model_engine_server.core.domain_exceptions` and `model_engine_server.domain.exceptions` though

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -635,8 +637,18 @@ def model_output_to_completion_output(
...
-                raise e
+                logger.exception(f"Error parsing text-generation-inference output {model_output}. Error message: {json.loads(e)['error']}")
+                '''
+                if 'generated_text' not in model_output:
+                    raise ObjectHasInvalidValueException(
+                        f"Error parsing text-generation-inference output {model_output}. Error message: {json.loads(e)['error']}"
+                    )
+                '''
+                if model_output.get("error_type") == "validation":
+                    raise InvalidRequestException(model_output.get("error")) # trigger a 400
```
yunfeng-scale, line 648: i think some changes are needed in llms_v1.py, are you able to get 400 repro validation error with the current changes?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -919,15 +931,23 @@ async def execute(
...
+                                finished=finished,
+                                num_completion_tokens=num_completion_tokens,
+                                token=token,
+                            ),
+                        )
+                    except Exception as e:  # if result["result"]["token"]["text"] doesn't exist 
+                        if "error" in result:
+                            logger.exception(f"Error parsing text-generation-inference stream output. Error message: {e}")
+                            raise ObjectHasInvalidValueException(
+                                f"Error parsing text-generation-inference stream output."
```
yunfeng-scale, line 948: i would expect same error processing logics here as in sync use case

model-engine/model_engine_server/core/domain_exceptions.py
ian-scale: oh, sorry about that -- I'll import from `model_engine_server.domain.exceptions`, I can't think of a reason why we'd need this in two places

ian-scale: Actually we should probably look into why we have two different sets of exceptions and combine them if possible. I'll look into that

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
ian-scale: updated to reflect that now

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
ian-scale: yeah, added them so that we handle and return 400 HTTPException

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -803,8 +813,10 @@ async def execute(
             ObjectNotAuthorizedException: If the owner does not own the model endpoint.
         """
 
+        logging.debug("do we reach execute()?")
```
yunfeng-scale, line 816: remove

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: raise InvalidRequestException or UpstreamServiceError?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -635,8 +642,12 @@ def model_output_to_completion_output(
                     tokens=tokens,
                 )
             except Exception as e:
-                logger.exception(f"Error parsing text-generation-inference output {model_output}")
-                raise e
+                logger.exception(f"Error parsing text-generation-inference output {model_output}.") 
+                if model_output.get("error_type") == "validation":
+                    raise InvalidRequestException(model_output.get("error")) # trigger a 400
+                else:
+                    raise UpstreamServiceError(model_output.get("error")) # also change llms_v1.py that will return a 500 HTTPException so user can retry
```
yunfeng-scale, line 649: can you double check 500s with proper error messages are returned in this code path?

ian-scale: I'm pretty sure that my most recent commit didn't get pushed, l don't see any of my changes to `llms_v1.py` showing up in the github UI. I'm really sorry for re-requesting a review pre-emptively. I'll fix these and then reach back out, sorry @yunfeng-scale

ian-scale:
error messages now look like this for `completions-sync` and `completions-stream`: 

Screenshot 2023-09-01 at 11 27 31 AM

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -230,8 +232,14 @@ async def create_completion_stream_task(
...
-            async for message in response:
-                yield {"data": message.json()}
+            try:
+                async for message in response:
+                    yield {"data": message.json()}
+            except InvalidRequestException as exc:
+                yield {
+                    "data": {"error": {"status_code": 400, "detail": str(exc)}}
+                }  # TODO: add pre-stream input validation with tokenizers
+                return
```
yunfeng-scale, line 242: can remove

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -919,15 +927,26 @@ async def execute(
...
+                            request_id=request_id,
+                            output=CompletionStreamOutput(
+                                text=result["result"]["token"]["text"],
+                                finished=finished,
+                                num_completion_tokens=num_completion_tokens,
+                                token=token,
+                            ),
+                        )
+                    except Exception as exc:
+                        logger.exception(f"Error parsing text-generation-inference output. {result['result']['error']}")
```
yunfeng-scale, line 941: i would log everything in result['result']

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -230,8 +232,14 @@ async def create_completion_stream_task(
         )
 
         async def event_generator():
-            async for message in response:
-                yield {"data": message.json()}
+            try:
+                async for message in response:
+                    yield {"data": message.json()}
+            except InvalidRequestException as exc:
+                yield {
```
yunfeng-scale, line 239: can you update docs about how to deal with errors when streaming

model-engine/model_engine_server/api/llms_v1.py
ian-scale: okay the TODO comment is removed

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
ian-scale: got it, made that fix!

model-engine/model_engine_server/api/llms_v1.py
ian-scale: yes, right now it looks like: 

```
stream = Completion.create(
    model="llama-2-7b",
    prompt="Give me a 200 word summary on the current economic events in the US.",
    max_new_tokens=1000,
    temperature=0.2,
    stream=True,
)

for response in stream:
  try:
      if response.output:
          print(response.output.text, end="")
          sys.stdout.flush()
  except: # an error occurred
    print(stream.text) # print the error message out 
```

yunfeng-scale: it sounds like you want to add them in a separate PR which is fine. be aware everything is in the same repo now

yunfeng-scale (approved):

model-engine/model_engine_server/api/llms_v1.py
ian-scale: I added them in this same pr

docs/guides/completions.md
```diff
@@ -85,9 +87,13 @@ stream = Completion.create(
...
 for response in stream:
-    if response.output:
-        print(response.output.text, end="")
-        sys.stdout.flush()
+    try:
+        if response.output:
+            print(response.output.text, end="")
+            sys.stdout.flush()
+    except: # an error occurred
+        print(stream.text) # print the error message out
```
yunfeng-scale, line 95: hmm, this doesn't seem right to me, shouldn't we print `response.error` or something? printing from `stream` doesn't feel right

PR #240, opened 2023-08-30
Add nvidia.com/gpu in requests

yunfeng-scale:
This tag should be in requests in addition to limits

### DIFF
--- a/charts/model-engine/templates/_helpers.tpl
+++ b/charts/model-engine/templates/_helpers.tpl
@@ -42,7 +42,7 @@ Common labels
 */}}
 {{- define "modelEngine.labels" -}}
 team: infra
-product: launch
+product: model-engine
 helm.sh/chart: {{ include "modelEngine.chart" . }}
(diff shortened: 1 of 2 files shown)


seanshi-scale (commented):
don't think we need the amount in requests for k8s, but if it's necessary for scripts to work then should be fine. Don't know if this was already discussed somewhere though.

https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/

yunfeng-scale:
> don't think we need the amount in requests for k8s, but if it's necessary for scripts to work then should be fine. Don't know if this was already discussed somewhere though.
> 
> https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/

that's what autoscaler used to calculate utilization, matters if we were to use scale-down-gpu-utilization-threshold https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/utilization/info.go#L122

seanshi-scale (approved):
oh ok, yeah that makes sense

PR #243, opened 2023-09-01
Validate Fine-tuning CSV headers

saiatmakuri:
check required headers "prompt" and "response" in the server before a job is started


model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
```diff
@@ -43,6 +46,23 @@ def ensure_model_name_is_valid_k8s_label(model_name: str):
     return re.sub("[^-A-Za-z0-9_.]", "-", model_name).lstrip("-_.")[:62].rstrip("-_.")
 
 
+def read_csv_headers(file_location: str):
+    """
+    Read the headers of a csv file. Assumes the file exists and is valid.
+    """
+    with smart_open.open(file_location) as file:
```
yunfeng-scale, line 53: can you try setting buffer size to be smaller? i think by default it prefetch 128KB data

yunfeng-scale (approved):

PR #244, opened 2023-09-01
Change max_input_length to half of max_total_tokens to work around potential tokenizer loading issue

seanshi-scale:


gargutsav (approved):

saiatmakuri (approved):

saiatmakuri (approved):

saiatmakuri (approved):

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -244,10 +244,10 @@ async def create_text_generation_inference_bundle(
         command = []
 
         # TGI requires max_input_length < max_total_tokens
-        max_input_length = 2047
+        max_input_length = 1024
```
yixu34, line 247: I think a comment would have been good here to explain why we're changing it. Useful context is that we're reducing `max_input_length` as a workaround for tokenizer issues, which we may fix at some point.

PR #248, opened 2023-09-01
Some fixes

yunfeng-scale:
1. don't by default use 4 GPUs when creating models
2. Fix balloon CPU deployment
3. Fix cacher datetime comparison

tested cache by running locally against prod passing comparison code

### DIFF
--- a/charts/model-engine/templates/balloon_cpu_deployment.yaml
+++ b/charts/model-engine/templates/balloon_cpu_deployment.yaml
@@ -34,7 +34,7 @@ spec:
           resources:
             limits:
               memory: 28Gi
-              cpu: 8
+              cpu: 6
           command:
             - /bin/bash
             - -c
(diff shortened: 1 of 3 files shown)


clients/python/llmengine/model.py
```diff
@@ -41,10 +41,10 @@ def create(
         quantize: Optional[Quantization] = None,
         checkpoint_path: Optional[str] = None,
         # General endpoint fields
-        cpus: int = 32,
-        memory: str = "192Gi",
+        cpus: int = 8,
+        memory: str = "40Gi",
```
yunfeng-scale, line 45: @gargutsav @auag92 i'm updating these default values, so if you're still going with 4 GPUs, please make sure values are set same as how you're using them right now

charts/model-engine/templates/balloon_cpu_deployment.yaml
```diff
@@ -34,7 +34,7 @@ spec:
           resources:
             limits:
               memory: 28Gi
-              cpu: 8
```
yunfeng-scale, line 37: m5.2xlarge has only 8 CPUs and about 1 is used for system cores

seanshi-scale (approved):

saiatmakuri (approved):

PR #249, opened 2023-09-01
Higher concurrency limit for gunicorn

yunfeng-scale:
Gunicorn returns 503 instead of 429 when concurrency exceeds the limit, before adding rate limiting just increase the concurrency and see if that's good enough

### DIFF
--- a/model-engine/model_engine_server/api/worker.py
+++ b/model-engine/model_engine_server/api/worker.py
@@ -1,8 +1,6 @@
 from uvicorn.workers import UvicornWorker
 
-# The target concurrency is around 50, so we set the limit to 32 with 4 workers
-# for a total concurrency of 128 to allow for some headroom.
-CONCURRENCY_LIMIT = 32
+CONCURRENCY_LIMIT = 1000
 
 
 class LaunchWorker(UvicornWorker):


ian-scale (commented):
is there any downside to increasing concurrency by such a large amount?

yunfeng-scale:
> is there any downside to increasing concurrency by such a large amount?

process might OOM, but there's an easy fix, just increase memory

ian-scale:
> > is there any downside to increasing concurrency by such a large amount?
> 
> process might OOM, but there's an easy fix, just increase memory

okay sounds good, approved!

ian-scale (approved):

model-engine/model_engine_server/api/worker.py
```diff
@@ -1,8 +1,6 @@
 from uvicorn.workers import UvicornWorker
 
-# The target concurrency is around 50, so we set the limit to 32 with 4 workers
-# for a total concurrency of 128 to allow for some headroom.
-CONCURRENCY_LIMIT = 32
+CONCURRENCY_LIMIT = 1000
```
yixu34, line 3: I would add a comment here, basically mirroring the PR description. If I saw this I would be pretty confused without context.

PR #254, opened 2023-09-02
Fix vllm docker tensor paralllel

yunfeng-scale:
there's some weird version conflict with pydantic using ray

tested by creating a llama 2 13b model with 2 shards


seanshi-scale (changes requested):
think we want to move out of `model-engine/` and delete it soon actually, since right now it's just a duplicate of `llm-engine` that isn't as up to date

yunfeng-scale:
> think we want to move out of `model-engine/` and delete it soon actually, since right now it's just a duplicate of `llm-engine` that isn't as up to date

oh hmm, how did i change things in that folder

yunfeng-scale:
> think we want to move out of `model-engine/` and delete it soon actually, since right now it's just a duplicate of `llm-engine` that isn't as up to date

wait this is indeed in model-engine. this folder is not models.

seanshi-scale (approved):
oh lol oops my bad

PR #255, opened 2023-09-05
Increase liveness timeout for main container

yunfeng-scale:
This came up during testing with vLLM. by default istio timeout is 1s for health check. vLLM docker sometimes can't respond within 1s during high load, making istio temporarily think the docker is unavailable and return 500s


seanshi-scale (approved):

charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -285,6 +285,12 @@ data:
                   port: ${USER_CONTAINER_PORT}
                 initialDelaySeconds: ${READINESS_INITIAL_DELAY}
                 periodSeconds: 5
+              livenessProbe:
+                httpGet:
+                  path: ${HEALTHCHECK_ROUTE}
+                  port: ${USER_CONTAINER_PORT}
+                initialDelaySeconds: ${READINESS_INITIAL_DELAY}
+                timeoutSeconds: 5
```
yixu34, line 293: Might be worth putting a comment saying that 5s helps for vLLM under load, per your PR description. Maybe same for the `readinessProbe`?

yixu34 (approved):
Anyway lgtm

PR #257, opened 2023-09-06
removing timezone tzinfo in favor of utc

ian-scale:


model-engine/model_engine_server/infra/services/image_cache_service.py
```diff
@@ -145,18 +144,18 @@ async def execute(self, endpoint_infra_states: Dict[str, Tuple[bool, ModelEndpoi
...
                             > images_to_cache_priority[key][state.image].last_updated_at
                         )
                         and self.docker_repository.image_exists(image_tag, repository_name)
                     ):
                         images_to_cache_priority[key][state.image] = cache_priority
 
         images_to_cache = CachedImages(cpu=[], a10=[], a100=[], t4=[])
+        for key, val in images_to_cache_priority.items():
+            for image in images_to_cache_priority[key]:
+                images_to_cache_priority[key][image].last_updated_at.replace(tzinfo=pytz.UTC)
```
yunfeng-scale, line 157: does .replace update the original object?

ian-scale: it doesn't look like it updates things inplace, changing it to be a new value assignment.

yunfeng-scale (approved):

PR #258, opened 2023-09-06
invalid CSV input returns InvalidRequestException

ian-scale:
Followup to @saiatmakuri's pr that added logic to check if `prompt, response` headers are valid. This is to check that the user passes in a valid file and surface an appropriate error message to them if they don't.

### DIFF
--- a/model-engine/customer_service_data.csv
+++ b/model-engine/customer_service_data.csv
@@ -1,0 +1,2 @@
+prompt,response
+What is your policy on carry-on luggage?,Our policy allows each passenger to bring one piece of carry-on luggage and one personal item such as a purse or briefcase. The maximum size for carry-on luggage is 22 x 14 x 9 inches.
--- a/model-engine/customer_service_data.csv
+++ b/model-engine/customer_service_data.csv
@@ -1,2 +1,0 @@
-prompt,response
-What is your policy on carry-on luggage?,Our policy allows each passenger to bring one piece of carry-on luggage and one personal item such as a purse or briefcase. The maximum size for carry-on luggage is 22 x 14 x 9 inches.
--- a/model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
@@ -48,11 +48,20 @@ def ensure_model_name_is_valid_k8s_label(model_name: str):
 
 def read_csv_headers(file_location: str):
     """
-    Read the headers of a csv file. Assumes the file exists and is valid.
+    Read the headers of a csv file. Raises a FileNotFoundError if the file does not exist. Raises a ValueError if csv is malformed.
     """
-    with smart_open.open(file_location, transport_params=dict(buffer_size=1024)) as file:
-        csv_reader = csv.DictReader(file)
-        return csv_reader.fieldnames
+    try:
+        with smart_open.open(file_location, transport_params=dict(buffer_size=1024)) as file:
+            csv_reader = csv.DictReader(file)
+            return csv_reader.fieldnames
+    except FileNotFoundError:
+        raise InvalidRequestException(
+            "The file that you provided cannot be found. Please check that the filepath is correct."
+        )
+    except csv.Error as exc:
+        raise InvalidRequestException(
+            f"An error occurred while processing the CSV file. Details: {exc}"
+        )
 
 
 def are_dataset_headers_valid(file_location: str):
--- a/model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
@@ -66,7 +66,7 @@ def are_dataset_headers_valid(file_location: str):
 
 def check_file_is_valid(file_name: Optional[str], file_type: str):
     """
-    Ensure the file is valid with required columns 'prompt' and 'response'.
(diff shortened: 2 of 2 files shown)


model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
```diff
@@ -48,11 +48,20 @@ def ensure_model_name_is_valid_k8s_label(model_name: str):
...
+    Read the headers of a csv file. Raises a FileNotFoundError if the file does not exist. Raises a ValueError if csv is malformed.
     """
-    with smart_open.open(file_location, transport_params=dict(buffer_size=1024)) as file:
-        csv_reader = csv.DictReader(file)
-        return csv_reader.fieldnames
+    try:
+        with smart_open.open(file_location, transport_params=dict(buffer_size=1024)) as file:
+            csv_reader = csv.DictReader(file)
+            return csv_reader.fieldnames
+    except FileNotFoundError:
```
saiatmakuri, line 57: two ideas:
- move try/except to `CreateFineTuneV1UseCase.execute()` to control exception handling in the main fn and distinguish errors between training/validation file more clearly
- create another "family" of validation helpers that raise all exceptions and move all the exception handling logic from `CreateFineTuneV1UseCase.execute` to these helper fns

model-engine/customer_service_data.csv
```diff
@@ -0,0 +1,2 @@
+prompt,response
```
saiatmakuri, line 1: is this used in a test case or documentation? if not, remove

saiatmakuri (approved):

model-engine/customer_service_data.csv
ian-scale: oh yeah this shouldn't have been committed

model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
ian-scale: Makes sense to move things into `CreateFineTuneV1UseCase.execute()`, I'll do that!

yixu34: +1 on moving the try/except here to the use case. Otherwise, if this helper needs to know about the use case, that's not ideal from a least privilege perspective. This helper should just know about reading files and how that can fail.

yixu34: > create another "family" of validation helpers that raise all exceptions and move all the exception handling logic from CreateFineTuneV1UseCase.execute to these helper fns

This can work too, though would just ask that any helpers be localized as much as possible. In other words, if we can continue to keep them in the use case file, instead of a global helpers/common/util file.

model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
```diff
@@ -63,6 +64,28 @@ def are_dataset_headers_valid(file_location: str):
...
 
+def check_file_is_valid(file_name: Optional[str], file_type: str):
+    """
+    Ensure the file is valid with required columns 'prompt' and 'response', isn't malformatted, and exists.
+    file_type: 'training' or 'validation'
+    """
+    try:
+        if file_name is not None and not are_dataset_headers_valid(file_name):
+            raise InvalidRequestException(
+                f"Required column headers {','.join(REQUIRED_COLUMNS)} not found in training dataset"
```
saiatmakuri, line 75: "training dataset" --> "{file_type} dataset"

model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
```diff
@@ -63,6 +64,28 @@ def are_dataset_headers_valid(file_location: str):
...
+            )
+        else:
+            return True
+    except FileNotFoundError:
+        raise InvalidRequestException(
+            f"The {file_type} file that you provided cannot be found. Please check that the filepath is correct."
```
saiatmakuri, line 81: nit: change message to "Cannot find the {file_type} file. Verify the path and file name are correct."

model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
```diff
@@ -63,6 +64,28 @@ def are_dataset_headers_valid(file_location: str):
...
+        )
+    except csv.Error as exc:
+        raise InvalidRequestException(
+            f"An error occurred while processing the {file_type} CSV file. Details: {exc}"
```
saiatmakuri, line 85: nit: change message to "Cannot parse the {file_type} dataset as CSV. Details: {exc}"

model-engine/model_engine_server/domain/use_cases/llm_fine_tuning_use_cases.py
```diff
@@ -63,6 +64,28 @@ def are_dataset_headers_valid(file_location: str):
...
+    Ensure the file is valid with required columns 'prompt' and 'response', isn't malformatted, and exists.
+    file_type: 'training' or 'validation'
+    """
+    try:
+        if file_name is not None and not are_dataset_headers_valid(file_name):
+            raise InvalidRequestException(
+                f"Required column headers {','.join(REQUIRED_COLUMNS)} not found in training dataset"
+            )
+        else:
+            return True
```
saiatmakuri, line 78: nit: can remove

PR #260, opened 2023-09-06
add timezone to max

ian-scale:
fix timezone part 2


yixu34 (approved):
Any tests to create or update?

yunfeng-scale:
can you first test this with private deployment to verifiy it works?

PR #261, opened 2023-09-07
Ianmacleod/add model delete

ian-scale:
delete model endpoints by name in llm-engine.


yunfeng-scale (commented):
add some unit tests?

model-engine/tests/unit/domain/test_llm_use_cases.py
```diff
@@ -869,3 +870,60 @@ async def test_download_nonexistent_model_raises_not_found(
...
+        model_endpoint_service=fake_model_endpoint_service,
+        llm_model_endpoint_service=fake_llm_model_endpoint_service,
+    )
+    user = User(user_id=test_api_key, team_id=test_api_key, is_privileged_user=True)
+    with pytest.raises(ObjectNotFoundException):
+        await use_case.execute(user=user, model_endpoint_name="nonexistent-model")
+
+
+@pytest.mark.asyncio
+async def test_delete_unauthorized_model_raises_not_authorized(
```
seanshi-scale, line 914: should we also have a test that a public model can't be deleted by anyone but the owner?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -779,8 +780,44 @@ async def execute(self, user: User, model_endpoint_name: str) -> GetLLMModelEndp
...
+            model_endpoint_name: The name of the model endpoint.
+
+        Returns:
+            A response object that contains a boolean indicating if deletion was successful.
+
+        Raises:
+            ObjectNotFoundException: If a model endpoint with the given name could not be found.
+            ObjectNotAuthorizedException: If the owner does not own the model endpoint.
+        """
+        model_endpoint = await self.llm_model_endpoint_service.get_llm_model_endpoint(
```
seanshi-scale, line 812: odd do we not pass a user object into here? don't know if we should be assuming model endpoint names are globally unique across all users

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -779,8 +780,44 @@ async def execute(self, user: User, model_endpoint_name: str) -> GetLLMModelEndp
...
+            model_endpoint_name
+        )
+        if not model_endpoint:
+            raise ObjectNotFoundException
+        if not self.authz_module.check_access_read_owned_entity(user, model_endpoint.record):
```
saiatmakuri, line 817: should we be using `check_access_write_owned_entity` here?

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -384,3 +388,41 @@ async def download_model_endpoint(
...
+        )
+        return await use_case.execute(user=auth, model_endpoint_name=model_endpoint_name)
+    except (ObjectNotFoundException) as exc:
+        raise HTTPException(
+            status_code=404,
+            detail="The requested model endpoint could not be found.",
+        ) from exc
+    except (ObjectNotAuthorizedException) as exc:
+        raise HTTPException(
+            status_code=401,
```
yixu34, line 416: Don't think this is correct, think we want a 403: https://authress.io/knowledge-base/articles/choosing-the-right-http-error-code-401-403-404

401 suggests that there's no auth; there _is_ auth, we're just denying the authenticated user.

Normally I'd actually suggest a 404 (think about how if you're not logged into Github and you go to a private repo you normally could see, it says 404 instead of 403). But since this is a public model name (?), maybe there's no point in being coy about the model existing or not.

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -384,3 +388,41 @@ async def download_model_endpoint(
...
+        ) from exc
+    except ExistingEndpointOperationInProgressException as exc:
+        raise HTTPException(
+            status_code=409,
+            detail="Existing operation on endpoint in progress, try again later.",
+        ) from exc
+    except EndpointDeleteFailedException as exc:  # pragma: no cover
+        raise HTTPException(
+            status_code=500,
+            detail="deletion of endpoint failed, compute resources still exist.",
```
yixu34, line 427: Don't think we need to or should mention details about compute resources existing.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -779,8 +780,44 @@ async def execute(self, user: User, model_endpoint_name: str) -> GetLLMModelEndp
...
+        model_endpoint_service: ModelEndpointService,
+        llm_model_endpoint_service: LLMModelEndpointService,
+    ):
+        self.model_endpoint_service = model_endpoint_service
+        self.llm_model_endpoint_service = llm_model_endpoint_service
+        self.authz_module = LiveAuthorizationModule()
+
+    async def execute(self, user: User, model_endpoint_name: str) -> DeleteLLMEndpointResponse:
+        """
+        Runs the use case to get the LLM endpoint with the given name.
```
yixu34, line 799: Get the LLM? This is delete right?

model-engine/model_engine_server/api/llms_v1.py
ian-scale: changed to 403.

model-engine/tests/unit/domain/test_llm_use_cases.py
```diff
@@ -869,3 +870,60 @@ async def test_download_nonexistent_model_raises_not_found(
...
+    fake_model_endpoint_service.add_model_endpoint(llm_model_endpoint_sync[0])
+    use_case = DeleteLLMEndpointByNameUseCase(
+        model_endpoint_service=fake_model_endpoint_service,
+        llm_model_endpoint_service=fake_llm_model_endpoint_service,
+    )
+    user = User(user_id=test_api_key, team_id=test_api_key, is_privileged_user=True)
+    response = await use_case.execute(
+        user=user, model_endpoint_name=llm_model_endpoint_sync[0].record.name
+    )
+    assert response.deleted is True
```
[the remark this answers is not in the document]
ian-scale: oh good idea, added that in

model-engine/tests/unit/domain/test_llm_use_cases.py
ian-scale: yes, will add now

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
ian-scale: 😅  fixing now

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
ian-scale: To incorporate user objects, I'll call `self.llm_model_endpoint_service.list_llm_model_endpoints` filtered by `name` and `owner`.

model-engine/tests/unit/domain/test_llm_use_cases.py
ian-scale: Actually, I think this test is actually subsumed by `test_delete_unauthorized_model` since no one but the owner would be authorized to delete any model - do you think there should be another separate test?

model-engine/tests/unit/conftest.py
```diff
@@ -3290,7 +3290,6 @@ def build_endpoint_request_async_custom(
         model_endpoint_record=ModelEndpointRecord(
             id="test_model_endpoint_id_3",
             name="test_model_endpoint_name_3",
-            created_by=test_api_key,
```
seanshi-scale, line 3293: why do we need to change this fixture? I'd imagine this may end up breaking various other tests

model-engine/tests/unit/domain/test_llm_use_cases.py
seanshi-scale: yeah, I was thinking that we'd probably add a test to potentially catch a case in the future where we change some authorization logic that ends up using the `public_model` field, and we inadvertently allow anyone to delete public models

model-engine/tests/unit/conftest.py
ian-scale: this doesn't need to change, I think this line was deleted accidentally -- undoing now

model-engine/tests/unit/domain/test_llm_use_cases.py
ian-scale: Where is the `public_model` field? Do you mean I should add a field called `public_model` to `ModelEndpointRecord`?

ian-scale: per discussion: ensure that there is a test that checks the `public_inference` field is True and when a user who has team access but not ownership of a model tries to delete it, the deletion fails.

ian-scale: I looked into this test more, and we currently determine write access based on the condition `return user.team_id == owned_entity.owner`, so either I should add a new authz method that checks that the `user.user_id == owned_entity.owner` or leave this as is and make the test with a different user_id and team_id than the owner of the endpoint. Currently, anyone who is under the same team_id as the initial model owner has the ability to delete the model. Is this not the desired end state?

seanshi-scale: It's fine to keep the check as `user.team_id == owned_entity.owner`, we've been assuming that people on the same team have full access to each other's resources (it's all owned by the team anyways), so we can keep it this way

saiatmakuri (approved):

PR #271, opened 2023-09-09
changing streaming error handling

ian-scale:


model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -238,6 +238,29 @@ async def event_generator():
...
+                return
+            except ObjectHasInvalidValueException as exc:
+                yield {"data": {"error": {"status_code": 400, "detail": str(exc)}}}
+                return
+            except EndpointUnsupportedInferenceTypeException as exc:
+                yield {
+                    "data": {
+                        "error": {
+                            "status_code": 400,
+                            "detail": f"Unsupported inference type: {str(exc)}",
```
yixu34, line 259: Would be ergonomic to tell the user what type they specified, and what the valid values are. Or does the inner exception tell them that?

yixu34 (commented):
Think we should have some API tests for this, as well as for the non-streaming path (if those don't already exist).

model-engine/model_engine_server/api/llms_v1.py
ian-scale: The inner error message tells them something more specific, i.e. 

`"Endpoint {model_endpoint_name} is not a streaming endpoint.`

https://github.com/scaleapi/llm-engine/blob/b8db3eab78ebd67e9c4f427b0a792e36598f24f1/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py#L1003-L1006

ian-scale:
> Think we should have some API tests for this, as well as for the non-streaming path (if those don't already exist).

looks like we skip these tests for streaming currently: 

https://github.com/scaleapi/llm-engine/blob/b8db3eab78ebd67e9c4f427b0a792e36598f24f1/model-engine/tests/unit/api/test_llms.py#L115-L121

model-engine/model_engine_server/api/llms_v1.py
yunfeng-scale: i think there errors are thrown at `use_case.execute` and not ever in `event_generator`?

ian-scale: oh man, I think I tested this incorrectly earlier, the error handling actually seems fine for the `use_case` level on ObjectNotFound and EndpointUnsupported. I'm gonna close this pr.

yunfeng-scale:
> > Think we should have some API tests for this, as well as for the non-streaming path (if those don't already exist).
> 
> looks like we skip these tests for streaming currently:
> 
> https://github.com/scaleapi/llm-engine/blob/b8db3eab78ebd67e9c4f427b0a792e36598f24f1/model-engine/tests/unit/api/test_llms.py#L115-L121
> 
> and we also don't have tests that test anything other than success for streaming or sync endpoints. Seems like we need to add quite a few more api tests, perhaps this would be better for a new PR?

when i was adding tests, can't get two streaming requests working in one test run.

PR #273, opened 2023-09-12
Integrate LightLLM

yunfeng-scale:
As title. stop sequences is not yet supported.

### DIFF
--- a/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
@@ -119,6 +119,15 @@ _SUPPORTED_MODEL_NAMES = {
         "falcon-40b": "tiiuae/falcon-40b",
         "falcon-40b-instruct": "tiiuae/falcon-40b-instruct",
     },
+    LLMInferenceFramework.LIGHTLLM: {
+        "llama-7b": "decapoda-research/llama-7b-hf",
+        "llama-2-7b": "meta-llama/Llama-2-7b-hf",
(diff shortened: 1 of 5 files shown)


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -499,6 +517,86 @@ async def create_vllm_bundle(
...
+        checkpoint_path: Optional[str],
+    ):
+        command = []
+
+        # TODO: incorporate auto calculate max_total_token_num from https://github.com/ModelTC/lightllm/pull/81
+        max_total_token_num = 6000  # LightLLM default
+        if num_shards == 1:
+            max_total_token_num = 15000  # Default for Llama 2 7B on 1 x A10
+        elif num_shards == 2:
+            max_total_token_num = 21000  # Default for Llama 2 13B on 2 x A10
```
ian-scale, line 536: just curious, why does the total number of tokens increase for 2 shards and then decrease for 3? is there any explanation for this (I expected it to be a linear trend)

yunfeng-scale: `max_total_token_num` is the full size of KV cache

yunfeng-scale: > why does the total number of tokens increase as shards go up

because we have more memory with more shards, also this is 100% a hack before automatic calculating `max_total_token_num`

ian-scale (approved):

PR #276, opened 2023-09-13
update docs

ian-scale:
updating the docs to reflect that there are multiple ways that users can set their API keys (especially useful for Jupyter notebook users so that they don't need to do any `os.environ` stuff).


yixu34 (approved):

yunfeng-scale (approved):

docs/getting_started.md
```diff
@@ -37,6 +37,12 @@ export SCALE_API_KEY="[Your API key]"
 
 You can also add in the line above to your `.zshrc` or `.bash_profile` so it's automatically set for future sessions.
 
+Alternatively, you can also set your API key using either of the following patterns:
```
yunfeng-scale, line 40: maybe being explicit that this is helpful in nodebook environment?

ian-scale: will add that in now

PR #278, opened 2023-09-15
fixing cacher, tested in prod version of cacher deployment in k8s and…

ian-scale:
… seems to be working

### DIFF
--- a/model-engine/model_engine_server/infra/services/image_cache_service.py
+++ b/model-engine/model_engine_server/infra/services/image_cache_service.py
@@ -64,7 +64,7 @@ class ImageCacheService:
             is_high_priority=1,  # make it a high priority
             has_no_available_workers=1,
             # assuming it has no available workers so that it will be at top after reverse sorting
-            last_updated_at=datetime.max,
+            last_updated_at=datetime.max.replace(tzinfo=pytz.utc),
             # setting it to max to ensure it will be at top after reverse sorting
         )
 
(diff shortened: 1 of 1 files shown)


seanshi-scale (approved):

model-engine/model_engine_server/infra/services/image_cache_service.py
```diff
@@ -162,9 +166,13 @@ async def execute(self, endpoint_infra_states: Dict[str, Tuple[bool, ModelEndpoi
...
-                val.keys(), key=lambda image: val[image], reverse=True
-            )[:IMAGES_TO_CACHE_PER_INSTANCE_TYPE]
+        try:
+            for key, val in images_to_cache_priority.items():
+                images_to_cache[key] = sorted(  # type: ignore
+                    val.keys(), key=lambda image: val[image], reverse=True
+                )[:IMAGES_TO_CACHE_PER_INSTANCE_TYPE]
+            logger.info("sorted images to cache successfully")
+        except Exception as exc:
+            logger.warning(f"sorting had an error. Error message: {exc}. Skipping sorting...")
```
saiatmakuri, line 176: with above changes, why would end up in a case with sorting broken? seems better to keep the integrity LRU cache than to just silently accept a _random_ cache if sorting fails

ian-scale: per discussion: will take out the logging and `try/except` block that catches errors here (if things are broken due to an unhandled error, they should be obviously broken so that we know they need fixing)

PR #280, opened 2023-09-18
add vllm to inference framework enum

ian-scale:
previously, due to this not being implemented on the client side we had a bug where `vllm` based endpoints would raise errors during `Model.get()` and `Model.list()`, eg: 

Screenshot 2023-09-18 at 10 21 21 AM

after this change we have: 
```
GetLLMEndpointResponse(..., status=<ModelEndpointStatus.READY: 'READY'>, inference_framework=<LLMInferenceFramework.VLLM: 'vllm'>...
```


yunfeng-scale (approved):

yunfeng-scale:
can you add lightllm as well?

PR #282, opened 2023-09-18
Ianmacleod/error handling

ian-scale:
Creates structured logging instead of having stack traces spread out over multiple logs. 
Consolidates exceptions into one unified place. 
Adds support for `request_id` that is surfaced when an uncaught exception occurs.


charts/model-engine/templates/_helpers.tpl
ian-scale, line 1: adding this configuration because we are currently seeing error logs related to this not being properly set up, the error logs go away when this is off. ref: https://github.com/DataDog/dd-trace-py/issues/5212

model-engine/model_engine_server/api/model_endpoints_v1.py
```diff
@@ -216,3 +243,10 @@ async def delete_model_endpoint(
             status_code=500,
             detail="deletion of endpoint failed, compute resources still exist.",
         ) from exc
+    except Exception as exc:
+        request_id = get_request_id()
```
yixu34, line 247: Think this is fine for now, but as we revisit self-hosting in VPCs, we'll probably want some sort of Gateway abstraction for getting the request/trace ID.

model-engine/model_engine_server/domain/exceptions.py
```diff
@@ -1,4 +1,62 @@
...
+    """
+
+
+@dataclass
+class DockerImageNotFoundException(DomainException):
+    """
+    Thrown when a user tries to specify a custom Docker image that cannot be found.
+    """
+
+    repository: str
```
yixu34, line 46: Just want to check that `repository` refers to a full image URL, minus the image tag. e.g. it shouldn't refer to an AWS ECR repo purely by its name. In other words, it shouldn't look like `some_service`, but `<account_id>.../some_service` or `<some dockerhub prefix>.../some_service`.

yixu34 (commented):
Just wanted to clarify the `repository` field

model-engine/model_engine_server/domain/exceptions.py
yixu34: Ok discussed offline, this is fine for now. If needed in the future, we can deprecate this field and come up with something new in its stead, like `image_url`.

yixu34 (approved):

PR #287, opened 2023-09-22
Ianmacleod/add datadog tracing in training

ian-scale:
trying to get context variables working so that datadog trace id is passed through the middleware. 

Current draft status: 

testing things out, still need to refactor and clean up the code after things start working 

error messages now look like this: 

```
Internal error for request_id bed5f209-5416-44da-93b0-18aeb27e79dc at time 2023-09-22 15:07:09 PDT. Our team has been notified.
```

### DIFF
--- a/model-engine/model_engine_server/api/app.py
+++ b/model-engine/model_engine_server/api/app.py
@@ -43,16 +43,13 @@ logger = make_logger(filename_wo_ext(__name__))
 
 
 class ExceptionLoggingMiddleware(BaseHTTPMiddleware):
-    def __init__(self):
-        self.timezone = pytz.timezone("US/Pacific")
-
     async def dispatch(self, request: Request, call_next):
         try:
             return await call_next(request)
         except Exception as e:
             tb_str = traceback.format_exception(etype=type(e), value=e, tb=e.__traceback__)
             request_id = get_request_id()
-            timestamp = datetime.now(self.timezone).strftime("%Y-%m-%d %H:%M:%S %Z")
+            timestamp = datetime.now(pytz.timezone("US/Pacific")).strftime("%Y-%m-%d %H:%M:%S %Z")
             add_trace_request_id(request_id)
             structured_log = {
                 "error": str(e),
--- a/model-engine/model_engine_server/api/app.py
+++ b/model-engine/model_engine_server/api/app.py
@@ -21,9 +21,15 @@ from model_engine_server.api.model_endpoints_docs_v1 import model_endpoints_docs
 from model_engine_server.api.model_endpoints_v1 import model_endpoint_router_v1
 from model_engine_server.api.tasks_v1 import inference_task_router_v1
 from model_engine_server.api.triggers_v1 import trigger_router_v1
-from model_engine_server.common.datadog_utils import add_trace_request_id, get_request_id
+from model_engine_server.common.datadog_utils import (
+    add_trace_request_id,
+    get_request_id,
+    request_id,
+    set_request_id_context,
+)
 from model_engine_server.core.loggers import filename_wo_ext, make_logger
-from starlette.middleware.base import BaseHTTPMiddleware
+
+logger = make_logger(filename_wo_ext(__name__))
 
 patch(fastapi=True)
 app = FastAPI(title="launch", version="1.0.0", redoc_url="/api")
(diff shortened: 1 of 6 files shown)


model-engine/model_engine_server/api/app.py
```diff
@@ -35,29 +45,31 @@
...
-            return JSONResponse(
-                {
-                    "status_code": 500,
-                    "content": {
-                        "error": f"Internal error for request_id {request_id}. Our team has been notified."
-                    },
-                }
-            )
 
+@app.middleware("http")
```
ian-scale, line 49: switched from ASGI middleware to fastapi style middleware in order to try to get DD trace ids to show up (didn't ultimately work, but both middlewares otherwise function the same way)

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -247,7 +249,7 @@ async def event_generator():
         request_id = get_request_id()
         logger.exception(f"Upstream service error for request {request_id}")
         return EventSourceResponse(
-            iter((CompletionStreamV1Response(request_id=request_id).json(),))
+            iter((CompletionStreamV1Response(request_id=request_id).json(),))  # type: ignore
```
ian-scale, line 252: had to add this type check since request_id can potentially be None.

model-engine/model_engine_server/api/model_endpoints_v1.py
```diff
@@ -92,13 +92,6 @@ async def create_model_endpoint(
             status_code=404,
             detail="The specified model bundle could not be found.",
         ) from exc
-    except Exception as exc:
```
ian-scale, line 95: removing generic exception handling here since it will be caught by the exception handling middleware.

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -133,6 +133,8 @@ async def list_model_endpoints(
     """
     Lists the LLM model endpoints owned by the current owner, plus all public_inference LLMs.
     """
+    # tracer.trace("tracer_test")
```
yunfeng-scale, line 136: remove these testing things?

ian-scale: removed now!

model-engine/model_engine_server/core/loggers.py
```diff
@@ -123,10 +119,14 @@ def make_json_logger(name: str, log_level: int = logging.INFO) -> logging.Logger
     in_kubernetes = os.getenv("KUBERNETES_SERVICE_HOST")
     if in_kubernetes:
         stream_handler.setFormatter(CustomJSONFormatter())
+    else:
+        stream_handler.setFormatter(CustomJSONFormatter())
```
yunfeng-scale, line 123: is this for testing? what about just remove if

model-engine/model_engine_server/api/app.py
```diff
@@ -18,10 +22,16 @@
...
+from model_engine_server.core.loggers import (
+    filename_wo_ext,
+    get_request_id,
+    make_logger,
+    set_request_id,
+)
+
+logger = make_logger(filename_wo_ext(__name__))
 
+patch(fastapi=True)
```
yixu34, line 34: Is this manual patching still necessary? I think you were adding this while testing, but hopefully this shouldn't be necessary.

model-engine/model_engine_server/core/loggers.py
```diff
@@ -123,10 +119,14 @@ def make_json_logger(name: str, log_level: int = logging.INFO) -> logging.Logger
...
+    """
```
yunfeng-scale, line 124: if setting to the same formatter remove the following comments?

model-engine/model_engine_server/api/app.py
yunfeng-scale, line 34: is this needed?

model-engine/model_engine_server/api/app.py
```diff
@@ -35,29 +45,31 @@
...
+            "error": str(e),
+            "request_id": str(request_id),
+            "traceback": "".join(tb_str),
+        }
+        logger.error("Unhandled exception: %s", structured_log)
+        return JSONResponse(
+            {
+                "status_code": 500,
+                "content": {
+                    "error": f"Internal error for request_id {request_id} at time {timestamp}. Our team has been notified."
```
yixu34, line 68: Think I'd rather just have the `error` be "Internal error occurred. Our team has been notified". And then have separate keys for `request_id` and `timestamp`.

yixu34: Also I thought we were going to have a DTO for this?

model-engine/model_engine_server/core/loggers.py
ian-scale: yeah this was for testing and can be removed, I think that we want to leave the logging as basic logging for testing locally with `just up` so that we don't have to read through the JSON logs.

yixu34, line 123: If we're doing this in both branches, can we just get rid of the `if` and `else`?

ian-scale: I actually think we should keep the if/else logic and use a basic logging config for local tests ( generally more readable, just had to have the JSON enriched version for testing the trace id / request id changes )

model-engine/model_engine_server/api/app.py
ian-scale: oh yeah I had previously written an InternalError DTO but then didn't ultimately use it, I can go back to using that.

ian-scale: [discussed offline] will actually leave it as JSONResponse since handling the new exception thrown by DTO is effectively the same.

ian-scale: new format of responses:
```
{"status_code":500,"content":{"error":"Internal error occurred. Our team has been notified.","timestamp":"2023-09-22 16:56:09 PDT","request_id":"60705bc0-8662-45fd-8b15-ecfbcb17cd8f"}}
```

model-engine/model_engine_server/api/app.py
ian-scale: just tested 1 more time and confirmed patching is not necessary

ian-scale: just tested 1 more time and confirmed patching is not necessary

yunfeng-scale (approved):

PR #290, opened 2023-09-25
Ianmacleod/pypi version nudge

ian-scale:
New behavior after this change: 

Screenshot 2023-09-25 at 3 27 25 PM

### DIFF
--- a/clients/python/llmengine/__init__.py
+++ b/clients/python/llmengine/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.0.b14"
+__version__ = "0.0.0.b13"
 
 import os
 from typing import Sequence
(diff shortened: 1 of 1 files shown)


clients/python/llmengine/__init__.py
```diff
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.0.beta13"
+__version__ = "0.0.0.b14"
```
seanshi-scale, line 15: wait did we forget to change this when upgrading from beta13 to beta14? oops if so

ian-scale: yeah oops :'(

ian-scale: Also, I am going to bump the version to 0.0.0b15 after this is approved so that everyone's client picks up these changes going forward.

yunfeng-scale (approved):

yixu34 (approved):

PR #291, opened 2023-09-26
Propagate extra server args to the gunicorn command

squeakymouse:


seanshi-scale (approved):

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -162,6 +162,7 @@ def entrypoint():
         str(args.num_workers),
         *envs,
         "model_engine_server.inference.forwarding.http_forwarder:app",
+        *extra_args,
```
seanshi-scale, line 165: oh wait hmm I think it might be good to just hardcode in the termination grace period seconds as a default here idk

seanshi-scale: `extra_args` feels like it should be a dict to me (e.g. `--key1 val1 --key2 val2` from `{key1: val1, key2: val2}`), is there such thing as `extra_kwargs`? 

this way we can set a default of graceful-timeout and override it if necessary

model-engine/model_engine_server/entrypoints/start_fastapi_server.py
```diff
@@ -8,7 +8,9 @@
 from typing import List
 
 
-def start_gunicorn_server(port: int, num_workers: int, debug: bool) -> None:
+def start_gunicorn_server(
```
seanshi-scale, line 11: wait this is the gateway right? I don't think we need to change things for the worker?

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
seanshi-scale: yeah it's fine to pass; I was hoping for a dict but if we don't get that from argparse then we can just directly pass it

PR #295, opened 2023-09-28
Add repetition_penalty, top_k, top_p to Completion

francesy-scale:
# Summary

Add the following parameters to Completion

* vllm, lightllm
  * presence_penalty
  * frequency_penalty
* text-generation-inference, vllm, lightllm
  * top_k
  * top_p

## Test Plan and Usage Guide

Test on local server

```
response = Completion.create(
    model="llama-2-7b-vllm",
    prompt="Im opening a pancake restaurant. List 3 quirky names I could name my restaurant.",
    max_new_tokens=100,
    temperature=0.6,
    frequency_penalty=0,
    presence_penalty=0,
    top_k=-1,
    top_p=0.6,
)
```
```
curl -X POST http://localhost:5001/v1/llm/completions-sync?model_endpoint_name=llama-2-7b-vllm -d '{"prompt":"Im opening a pancake restaurant. List 3 quirky names I could name my restaurant.", "max_new_tokens":100, "temperature":0.6, "presence_penalty":0, "frequency_penalty":0, "top_k":-1, "top_p":0.6}' -H "content-type: application/json"
```
"\nThe name of my restaurant is \"The Pancake House\".\nI'm opening a pancake restaurant. List 3 quirky names I could name my restaurant.\nI'm opening a pancake restaurant. List 3 quirky names I could name my restaurant.?\nI'm opening a pancake restaurant. List 3 quirky names I could name my restaurant.? I'm opening a pancake restaurant. List ",

```
response = Completion.create(
    model="llama-2-7b-vllm",
    prompt="Im opening a pancake restaurant. List 3 quirky names I could name my restaurant.",
    max_new_tokens=100,
    temperature=0.6,
    frequency_penalty=2,
    presence_penalty=2,
    top_k=-1,
    top_p=0.6,
)
```
```
curl -X POST http://localhost:5001/v1/llm/completions-sync?model_endpoint_name=llama-2-7b-vllm -d '{"prompt":"Im opening a pancake restaurant. List 3 quirky names I could name my restaurant.", "max_new_tokens":100, "temperature":0.6, "presence_penalty":2, "frequency_penalty":2, "top_k":-1, "top_p":0.6}' -H "content-type: application/json"
```
"\nI'm opening a pancake restaurant. List 3 quirky names I could name my restaurant.\nYou can use the following ideas to get you started:\nThe Pancake Shack (or Hut) - This is a simple and straightforward name that will appeal to most people, but it may not be as memorable or unique as some of the other options on this list. If you want something more creative, consider using one of these alternatives instead: The"


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -937,6 +937,8 @@ async def execute(
 
         request_id = str(uuid4())
         add_trace_request_id(request_id)
+        if request.top_k == 0:  # top_k can't be 0, only takes >= 1, or -1/None to disable top_k
+            request.top_k = -1
```
adlam-scale, line 941: can you add a check for validity of vllm/tgi/lightllm specific parameters? Otherwise the args will pass through silently and users will not know what happened.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1222,21 +1245,35 @@ async def execute(
...
                     "max_new_tokens": request.max_new_tokens,
+                    "repetition_penalty": request.repetition_penalty,
                 },
             }
             if request.stop_sequences is not None:
                 args["parameters"]["stop"] = request.stop_sequences
             if request.temperature > 0:
                 args["parameters"]["temperature"] = request.temperature
                 args["parameters"]["do_sample"] = True
+                if request.top_k == -1:  # tgi set to None to consider all tokens.
```
yixu34, line 1256: Ah alternatively, you can keep these as `Optional` like you have, but default to `None`.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yixu34: +1, also we should centralize all of the framework-specific validation, like later in this function.

yixu34 (commented):
This is not specific to this PR, but I think we might be at the point where we have centralized framework validation modules for TGI, vLLM, LightLLM, etc. We're currently duplicating that logic in the streaming and non-streaming paths. @yunfeng-scale @ian-scale thoughts?

clients/python/llmengine/completion.py
```diff
@@ -72,6 +77,32 @@ async def acreate(
...
+            presence_penalty (Optional[float]):
+                Only affects: vllm, lightllm
+                Penalize new tokens based on whether they appear in the text so far, increasing the model's likelihood to talk about new topics.
+                https://platform.openai.com/docs/guides/gpt/parameter-details
+                Range: [0.0, 2.0]. Higher values encourage the model to use new tokens.
+
+            frequency_penalty (Optional[float]):
+                Only affects: vllm, lightllm
+                Penalize new tokens based on their existing frequency in the text so far, decreasing the model's likelihood to repeat the same line verbatim.
+                https://platform.openai.com/docs/guides/gpt/parameter-details
```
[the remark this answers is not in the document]
francesy-scale: vllm didn't link sources for them, but google search result says presence_penalty and frequency_penalty are from OpenAI, and the range matches too, so I linked it here

clients/python/llmengine/completion.py
```diff
@@ -72,6 +77,32 @@ async def acreate(
                 Whether to return the log probabilities of generated tokens.
                 When True, the response will include a list of tokens and their log probabilities.
 
+            repetition_penalty (Optional[float]):
```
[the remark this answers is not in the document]
francesy-scale: I checked, they have the same intention, but different implementation. The range is also different, `repetition_penalty` takes [1, infinity), `frequency_penalty` takes [-2, 2]. I thought it will be confusing if we simply replace the name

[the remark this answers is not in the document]
yunfeng-scale: let me take a look. i think exposing all these parameters would be confusing to end users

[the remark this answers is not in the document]
yunfeng-scale: looks like TGI's penalty uses division like in https://arxiv.org/pdf/1909.05858.pdf and vLLM and lightLLM uses minus. let's remove `repetition_penalty` and provide only `presence_penalty` and `frequency_penalty`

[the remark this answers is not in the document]
francesy-scale: TGI only supports `repetition_penalty`. Does this mean if users choose TGI framework, they won't have to option to penalize repetition?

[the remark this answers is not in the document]
yunfeng-scale: yes, i would suggest anyone wants to use repetition penalty to migrate to vLLM.

clients/python/llmengine/completion.py
```diff
@@ -33,6 +33,10 @@ async def acreate(
         temperature: float = 0.2,
         stop_sequences: Optional[List[str]] = None,
         return_token_log_probs: Optional[bool] = False,
+        presence_penalty: float = 0.0,  # vllm, lightllm
```
yunfeng-scale, line 36: can you make these optional? is there a reason for them to be not optional?

clients/python/llmengine/completion.py
```diff
@@ -195,6 +227,10 @@ def create(
...
```
yunfeng-scale, line 230: make optional

clients/python/llmengine/completion.py
yunfeng-scale: also no need to comment here about frameworks since it's specific later in main comment

clients/python/llmengine/data_types.py
```diff
@@ -268,6 +268,10 @@ class CompletionSyncV1Request(BaseModel):
     temperature: float = Field(..., ge=0.0)
     stop_sequences: Optional[List[str]] = Field(default=None)
     return_token_log_probs: Optional[bool] = Field(default=False)
+    presence_penalty: float = Field(default=0.0, ge=0.0, le=2.0)
```
yunfeng-scale, line 271: make optional

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -831,6 +831,36 @@ def deepspeed_result_to_tokens(result: Dict[str, Any]) -> List[TokenOutput]:
     return tokens
 
 
+def validate_completion_params(
```
yunfeng-scale, line 834: can you rename it to validate_and_update_completion_params since this is not a readonly function?

yunfeng-scale (approved):
some nits. please address before merge, thanks!

clients/python/llmengine/completion.py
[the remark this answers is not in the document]
yunfeng-scale: there's no need but i don't see why we want to have them as non-optional

yunfeng-scale:
> This is not specific to this PR, but I think we might be at the point where we have centralized framework validation modules for TGI, vLLM, LightLLM, etc. We're currently duplicating that logic in the streaming and non-streaming paths. @yunfeng-scale @ian-scale thoughts?

yeah some refactoring (probably plus unit tests) is needed

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -831,6 +831,54 @@ def deepspeed_result_to_tokens(result: Dict[str, Any]) -> List[TokenOutput]:
     return tokens
 
 
+def validate_and_update_completion_params(
+    inference_framework: LLMInferenceFramework,
+    request: Union[CompletionSyncV1Request, CompletionStreamV1Request],
```
seanshi-scale, line 836: btw if the type checker is still giving you trouble maybe https://docs.python.org/3.8/library/typing.html#user-defined-generic-types would help? at least this feels like the "proper" way to do things to me

francesy-scale: I was able to pass the check if I call the function like this
```
new_request = validate_and_update_completion_params(endpoint_content.inference_framework, request)
assert isinstance(new_request, CompletionSyncV1Request)
request = new_request
```

seanshi-scale: does something like `new_request: CompletionSyncV1Request = validate...` also work? basically another way of telling the type checker that you know it'll be a CompletionSyncV1Request

francesy-scale: Doesn't work... I think the problem is if the return type is union, will have to do some type narrowing https://mypy.readthedocs.io/en/stable/type_narrowing.html

yunfeng-scale: you can use generics https://mypy.readthedocs.io/en/stable/generics.html

```
from typing import TypeVar, Sequence

T = TypeVar('T')

def validate_and_update_completion_params(
    inference_framework: LLMInferenceFramework,
    request: T,
):
...
```

jenkspt:
I just tried
```python
from llmengine import Completion

response = Completion.create(
    model="llama-2-7b",
    prompt="Hello, my name is",
    max_new_tokens=10,
    temperature=0.2,
    top_p=0.6,
)

print(response.json())
print(response.output.text)
```
And getting
```
      1 from llmengine import Completion
----> 3 response = Completion.create(
      4     model="llama-2-7b",
      5     prompt="Hello, my name is",
      6     max_new_tokens=10,
      7     temperature=0.2,
      8     top_p=0.6,
      9 )
     11 print(response.json())
     12 # '{"request_id": "c4bf0732-08e0-48a8-8b44-dfe8d4702fb0", "output": {"text": "________ and I am a ________", "num_completion_tokens": 10}}'

TypeError: create() got an unexpected keyword argument 'top_p'
I see top_p in the python [client documentation](https://llm-engine.scale.com/api/python_client/#:~:text=responses%20or%20not.-,create,-classmethod)
```
I see `top_p` in the python [client documentation](https://llm-engine.scale.com/api/python_client/#:~:text=responses%20or%20not.-,create,-classmethod)

Do we need to add some tests for these parameters?

francesy-scale:
> Do we need to add some tests for these parameters?

`pip install scale-llm-engine --upgrade` should fix the issue now. I will add some unit tests

PR #296, opened 2023-09-28
Smartly check safetensors vs. bin

sam-scale:
Tested via:

```
(llm-engine) (🚂 finetune_pipeline) ubuntu@ip-172-26-56-82:~/Code/llm-engine/model-engine$ python
Python 3.8.17 | packaged by conda-forge | (default, Jun 16 2023, 07:06:00)
[GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from model_engine_server.core.aws.storage_client import s3_list_files
>>> checkpoint_path = "s3://scale-ml/federal/donovan/checkpoints/llama-2-13b-chat/llama-2-13b-chat-2023-09-27_00:07:45.729686/checkpoint-64-compatible/"
>>> from model_engine_server.core.utils.url import parse_attachment_url
>>> parsed_remote = parse_attachment_url(checkpoint_path)
>>> all_files = s3_list_files(bucket=parsed_remote.bucket, key=parsed_remote.key)
2023-09-28 18:19:51,252 INFO [config] [config.py:54] - Using config file path: `/home/ubuntu/Code/llm-engine/model-engine/model_engine_server/core/configs/default.yaml`
>>> all_files[0]
'federal/donovan/checkpoints/llama-2-13b-chat/llama-2-13b-chat-2023-09-27_00:07:45.729686/checkpoint-64-compatible/added_tokens.json'
>>> len([f for f in all_files if f.endswith(".safetensors")])
0
>>> len([f for f in all_files if f.endswith(".bin")])
6
>>> exclude_str = "*.bin"
>>> exclude_str = "*.safetensors"
>>> s5cmd = "./s5cmd"
>>> import os
>>> final_weights_folder = ''
>>> print(f"{s5cmd} --numworkers 512 cp --concurrency 10 --exclude '{exclude_str}' {os.path.join(checkpoint_path, '*')} {final_weights_folder}")
./s5cmd --numworkers 512 cp --concurrency 10 --exclude '*.safetensors' s3://scale-ml/federal/donovan/checkpoints/llama-2-13b-chat/llama-2-13b-chat-2023-09-27_00:07:45.729686/checkpoint-64-compatible/*
>>> exit()
```


model-engine/model_engine_server/core/aws/storage_client.py
```diff
@@ -1,5 +1,5 @@
 import time
```
ian-scale, line 1: Can you run the pre-commit hooks? Looks like the unit tests are currently failing on black checks. Otherwise looks good!

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -358,14 +360,22 @@ def load_model_weights_sub_commands(
...
         else:
-            if framework == LLMInferenceFramework.TEXT_GENERATION_INFERENCE:
-                subcommands.append(
-                    f"{s5cmd} --numworkers 512 cp --concurrency 10 {os.path.join(checkpoint_path, '*')} {final_weights_folder}"
-                )
+            # Let's check whether to exclude "*.safetensors" or "*.bin" files
+            parsed_remote = parse_attachment_url(checkpoint_path)
+            all_files = s3_list_files(bucket=parsed_remote.bucket, key=parsed_remote.key)
+
+            # If there are more files ending in .safetensors, then exclude *.bin
```
ian-scale, line 367: Also, is there a way to add a couple unit tests to make sure that this works? I think it would be great to add those if we can.

sam-scale: Yep let me take a look at how that could make sense

model-engine/tests/unit/domain/test_llm_use_cases.py
```diff
@@ -953,3 +958,53 @@ async def test_delete_public_inference_model_raises_not_authorized(
...
+)
+@pytest.mark.parametrize("image_tag", ["0.9.3-launch_s3"])
+@pytest.mark.parametrize(
+    "checkpoint_path",
+    [
+        "s3://scale-ml/testing/llama-2-13b-bin-path/",
+        "s3://scale-ml/testing/llama-2-13b-safetensors-path/",
+    ],
+)
+def test_load_model_weights_sub_commands(
```
ian-scale, line 980: I should've been more specific about the unit tests. It looks like this unit test calls the live use case and endpoint service instead of using the fake endpoint service pytest fixture. I think that this is fine, but we might want to treat it more like an integration test that runs against our live infrastructure instead of a unit test (since it implicitly relies on other components working for this test to pass).

model-engine/tests/unit/domain/test_llm_use_cases.py
```diff
@@ -953,3 +958,53 @@ async def test_delete_public_inference_model_raises_not_authorized(
...
+        LLMInferenceFramework.LIGHTLLM,
+        LLMInferenceFramework.TEXT_GENERATION_INFERENCE,
+        LLMInferenceFramework.VLLM,
+    ],
+)
+@pytest.mark.parametrize("image_tag", ["0.9.3-launch_s3"])
+@pytest.mark.parametrize(
+    "checkpoint_path",
+    [
+        "s3://scale-ml/testing/llama-2-13b-bin-path/",
```
ian-scale, line 976: Are we actually accessing this bucket every time we run this unit test? If so, I think that means we need to run with appropriate AWS permissions, maybe we should create a fake s3 bucket object for bin-path and safetensors-path so that we don't need to do this.

model-engine/tests/unit/domain/test_llm_use_cases.py
```diff
@@ -953,3 +958,53 @@ async def test_delete_public_inference_model_raises_not_authorized(
...
+@pytest.mark.parametrize(
+    "framework",
+    [
+        LLMInferenceFramework.DEEPSPEED,
+        LLMInferenceFramework.LIGHTLLM,
+        LLMInferenceFramework.TEXT_GENERATION_INFERENCE,
+        LLMInferenceFramework.VLLM,
+    ],
+)
+@pytest.mark.parametrize("image_tag", ["0.9.3-launch_s3"])
```
ian-scale, line 972: Why do we need this?

model-engine/tests/unit/domain/test_llm_use_cases.py
ian-scale: @sam-scale okay we have to remove this for sure because we can't mention our s3 bucket location in the public repo.

sam-scale (commented):
lg!

yunfeng-scale (approved):

model-engine/tests/unit/domain/test_llm_use_cases.py
```diff
@@ -953,3 +960,15 @@ async def test_delete_public_inference_model_raises_not_authorized(
         await use_case.execute(
             user=user, model_endpoint_name=llm_model_endpoint_sync[0].record.name
         )
+
+
+@pytest.mark.asyncio
+async def test_exclude_safetensors_or_bin_majority_bin_works():
+    fake_model_files = ["fake.bin", "fake2.bin", "fake3.safetensors"]
```
yunfeng-scale, line 967: can you add some other files like `model.json`

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -358,14 +375,16 @@ def load_model_weights_sub_commands(
...
-                )
-            else:
-                subcommands.append(
-                    f"{s5cmd} --numworkers 512 cp --concurrency 10 --exclude '*.safetensors'  {os.path.join(checkpoint_path, '*')} {final_weights_folder}"
-                )
+            # Let's check whether to exclude "*.safetensors" or "*.bin" files
+            checkpoint_files = self.llm_artifact_gateway.get_files_from_checkpoint(checkpoint_path)
+            model_files = [f for f in checkpoint_files if "model" in f]
+
+            exclude_str = _exclude_safetensors_or_bin(model_files)
```
sam-scale, line 382: Syncing offline, we should have this function return the string of: 
```
--exclude '*.bin'
```

because in the case of a tie we want it to return ''

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -134,6 +134,21 @@
...
+def _exclude_safetensors_or_bin(model_files):
+    """
+    This function is used to determine whether to exclude "*.safetensors" or "*.bin" files
+    based on which file type is present more often in the checkpoint folder.
+    """
+    exclude_str = ""
+    if len([f for f in model_files if f.endswith(".safetensors")]) > len(
+        [f for f in model_files if f.endswith(".bin")]
+    ):
+        exclude_str = "*.bin"
```
sam-scale, line 146: change to '--exclude '*.bin'"

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -134,6 +134,21 @@
...
+    else:
+        exclude_str = "*.safetensors"
```
sam-scale, line 148: change to '--exclude '*.bin'"

model-engine/model_engine_server/infra/gateways/s3_llm_artifact_gateway.py
```diff
@@ -17,6 +18,21 @@ def _get_s3_resource(self, kwargs):
         resource = session.resource("s3")
         return resource
 
+    def get_files_from_checkpoint(self, checkpoint_path: str, **kwargs) -> List[str]:
```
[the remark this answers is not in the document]
ian-scale: sure, sounds good -- the specific part is that I'm passing in a url for the checkpoint path in the function definition

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
ian-scale: Opted to instead keep the return statement as is and add some logic around the subcommands.append() call, I think it should have the same effect.

model-engine/tests/unit/domain/test_llm_use_cases.py
```diff
@@ -953,3 +960,34 @@ async def test_delete_public_inference_model_raises_not_authorized(
         await use_case.execute(
             user=user, model_endpoint_name=llm_model_endpoint_sync[0].record.name
         )
+
+
+@pytest.mark.asyncio
+async def test_exclude_safetensors_or_bin_majority_bin_works():
```
yixu34, line 966: nit: would generally suggest something more descriptive than "works"

model-engine/model_engine_server/infra/gateways/s3_llm_artifact_gateway.py
```diff
@@ -17,6 +18,21 @@ def _get_s3_resource(self, kwargs):
...
+        bucket = parsed_remote.bucket
+        key = parsed_remote.key
+        try:
+            # From here: https://dev.to/aws-builders/how-to-list-contents-of-s3-bucket-using-boto3-python-47mm
+            files = [
+                bucket_object["Key"]
+                for bucket_object in s3.list_objects_v2(Bucket=bucket, Prefix=key)["Contents"]
+            ]
+        except Exception as e:  # type: ignore
+            raise e
```
yixu34, line 33: This doesn't seem to do anything? We're just catching the exception and raising it again?

ian-scale: 😅 taken out

model-engine/tests/unit/domain/test_llm_use_cases.py
ian-scale: fair point, will do

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -134,6 +134,21 @@
 DOWNSTREAM_REQUEST_TIMEOUT_SECONDS = 5 * 60  # 5 minutes
 
 
+def _exclude_safetensors_or_bin(model_files):
+    """
+    This function is used to determine whether to exclude "*.safetensors" or "*.bin" files
```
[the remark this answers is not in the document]
sam-scale: I'm open to other suggestion, but given that we take as input:
* list of files

And output:
* which type of files

I think this feels like the right way to make the choice. I haven't ever seen a model saved in such a way where there are more tensors than bin files and you want it loaded via bin files (or vice versa)

The only thing I can think of is if somehow you save some separate model in the same directory as `.bin` and save your LLM as `.safetensors`, but that feels like a reasonable thing to ask of the user.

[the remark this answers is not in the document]
yixu34: Yeah I'm fine with what y'all decide here, would just ask that we have that logic updated in our doc site.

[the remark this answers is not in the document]
ian-scale: I'm going to merge this in now to address @victor-gz 's immediate concern but would be happy to update the docs as well to make the fact that we're doing this explicit.

PR #297, opened 2023-09-29
Add A100e as GPU type

yunfeng-scale:


clients/python/llmengine/data_types.py
```diff
@@ -32,6 +32,7 @@ class GpuType(str, Enum):
     NVIDIA_TESLA_T4 = "nvidia-tesla-t4"
     NVIDIA_AMPERE_A10 = "nvidia-ampere-a10"
     NVIDIA_AMPERE_A100 = "nvidia-ampere-a100"
+    NVIDIA_AMPERE_A100E = "nvidia-ampere-a100e"
```
ian-scale, line 35: can you check if this also needs to change on the spellbook side?

ian-scale (approved):

clients/python/llmengine/data_types.py
ian-scale: > can you check if this also needs to change on the spellbook side?

asking because I know that was the case when we made changes for the inference framework

yixu34 (approved):

clients/python/llmengine/data_types.py
yunfeng-scale: i think spellbook won't need a change since this field will not be populated

yunfeng-scale:
looks like checks won't run if they're from a fork

PR #302, opened 2023-10-02
adding s3 session function instead of client function

ian-scale:
to fix bug with Model.create().


ian-scale:
This was tested locally with logging statements while creating new `llama-2-7b` endpoints to confirm that the behavior of `Model.create()` now works.

sam-scale (approved):

PR #306, opened 2023-10-03
Increase graceful timeout and hardcode AWS_PROFILE

squeakymouse:


model-engine/model_engine_server/infra/gateways/resources/k8s_resource_types.py
```diff
@@ -507,6 +507,7 @@ def get_endpoint_resource_arguments_from_request(
     main_env = []
     if isinstance(flavor, RunnableImageLike) and flavor.env:
         main_env = [{"name": key, "value": value} for key, value in flavor.env.items()]
+    main_env.append({"name": "AWS_PROFILE", "value": build_endpoint_request.aws_role})
```
seanshi-scale, line 510: should we put a test for this (to test various merging logic with user-provided aws profiles)

model-engine/model_engine_server/inference/sync_inference/start_fastapi_server.py
```diff
@@ -10,7 +10,7 @@
 
 def start_server():
     parser = argparse.ArgumentParser()
-    parser.add_argument("--graceful-timeout", type=int, default=600)
+    parser.add_argument("--graceful-timeout", type=int, default=1800)
```
yunfeng-scale, line 13: is this what's being used in async tasks?

yunfeng-scale: i thought we are going to patch llm-engine/model-engine/model_engine_server/inference/async_inference/celery.py to listen to sigterm?

seanshi-scale: yeah iirc this code is used to serve the user container for both sync and async tasks for artifact-like bundles, and `async_inference/celery.py` shouldn't be used anymore

seanshi-scale: we'd also have to patch celery-forwarder to listen to sigterm

seanshi-scale (approved):

model-engine/model_engine_server/inference/sync_inference/start_fastapi_server.py
yunfeng-scale: possible to simulate such an scenario by sending traffic while restarting a pod?

PR #307, opened 2023-10-03
Ianmacleod/add mistral

ian-scale:
adding mistral endpoint support for completions

### DIFF
--- a/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
@@ -117,6 +117,7 @@ _SUPPORTED_MODEL_NAMES = {
         "falcon-7b-instruct": "tiiuae/falcon-7b-instruct",
         "falcon-40b": "tiiuae/falcon-40b",
         "falcon-40b-instruct": "tiiuae/falcon-40b-instruct",
+        "mistral-7b-instruct": "mistralai/Mistral-7B-Instruct-v0.1",
     },
     LLMInferenceFramework.LIGHTLLM: {
         "llama-7b": "decapoda-research/llama-7b-hf",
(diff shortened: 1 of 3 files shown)


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -117,6 +117,8 @@
         "falcon-7b-instruct": "tiiuae/falcon-7b-instruct",
         "falcon-40b": "tiiuae/falcon-40b",
         "falcon-40b-instruct": "tiiuae/falcon-40b-instruct",
+        "mistral-7b": "mistralai/Mistral-7B-v0.1",
```
yixu34, line 120: Do we need to add additional validation/tests? Also, think we need to update the EGP APIs?

Also we may want to update our Model Zoo docs to have some more granularity, to reflect the fact that not all models are supported for all frameworks.

yixu34, line 120: Do we need to add additional validation/tests? Also, think we need to update the EGP APIs?

Also we may want to update our Model Zoo docs to have some more granularity, to reflect the fact that not all models are supported for all frameworks.

yixu34 (approved):

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
ian-scale: Oh yes, the EGP APIs will also need to change. And I will add the model-zoo comment and add a supported frameworks column to the docs.

ian-scale: as far as testing, we can add synthetic monitors for completions for the new models. I will take out a new pr for this.

yunfeng-scale, line 120: update L494 about context window?

PR #311, opened 2023-10-09
Adding code llama to TGI

mfagundo-scale:
Adding code llama to endpoints supported with TGI.

Code Llama is a Llama architecture, and thus supported for optimization with TGI.


ian-scale (approved):
looks good to me!

PR #314, opened 2023-10-10
Fix streaming endpoint failure handling

yunfeng-scale:
- Fix streaming endpoint failure handling

before the fix, getting error:
```
urllib3.exceptions.ProtocolError: ("Connection broken: InvalidChunkLength(got length b'', 0 bytes read)", InvalidChunkLength(got length b'', 0 bytes read))
```

after the fix, getting response:
```
b"data: {'error': {'status_code': 404, 'detail': 'Model endpoint a not found.'}}\r\n\r\n"
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,6 +55,7 @@ repos:
     hooks:
       - id: mypy
         name: mypy-clients-python
+        files: clients/python/.*.py
         entry: mypy --config-file clients/python/mypy.ini
         language: system
   - repo: https://github.com/pre-commit/mirrors-mypy
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,7 +55,7 @@ repos:
     hooks:
       - id: mypy
         name: mypy-clients-python
-        files: clients/python/.*.py
+        files: clients/python/.*
         entry: mypy --config-file clients/python/mypy.ini
         language: system
   - repo: https://github.com/pre-commit/mirrors-mypy
--- a/model-engine/model_engine_server/api/llms_v1.py
+++ b/model-engine/model_engine_server/api/llms_v1.py
@@ -226,42 +226,38 @@ async def create_completion_stream_task(
     logger.info(
         f"POST /completion_stream with {request} to endpoint {model_endpoint_name} for {auth}"
     )
-    try:
-        use_case = CompletionStreamV1UseCase(
-            model_endpoint_service=external_interfaces.model_endpoint_service,
-            llm_model_endpoint_service=external_interfaces.llm_model_endpoint_service,
-        )
-        response = use_case.execute(
-            user=auth, model_endpoint_name=model_endpoint_name, request=request
-        )
+    use_case = CompletionStreamV1UseCase(
+        model_endpoint_service=external_interfaces.model_endpoint_service,
+        llm_model_endpoint_service=external_interfaces.llm_model_endpoint_service,
+    )
+    response = use_case.execute(user=auth, model_endpoint_name=model_endpoint_name, request=request)
 
-        async def event_generator():
-            try:
-                async for message in response:
-                    yield {"data": message.json()}
-            except InvalidRequestException as exc:
-                yield {"data": {"error": {"status_code": 400, "detail": str(exc)}}}
-                return
+    async def event_generator():
+        try:
+            async for message in response:
+                yield {"data": message.json()}
+        except InvalidRequestException as exc:
+            yield {"data": {"error": {"status_code": 400, "detail": str(exc)}}}
+        except UpstreamServiceError as exc:
+            request_id = get_request_id()
+            logger.exception(f"Upstream service error for request {request_id}")
+            yield {"data": {"error": {"status_code": 500, "detail": str(exc)}}}
+        except (ObjectNotFoundException, ObjectNotAuthorizedException) as exc:
+            print(str(exc))
+            yield {"data": {"error": {"status_code": 404, "detail": str(exc)}}}
+        except ObjectHasInvalidValueException as exc:
+            yield {"data": {"error": {"status_code": 400, "detail": str(exc)}}}
+        except EndpointUnsupportedInferenceTypeException as exc:
+            yield {
+                "data": {
+                    "error": {
+                        "status_code": 400,
+                        "detail": f"Unsupported inference type: {str(exc)}",
+                    }
+                }
+            }
 
-        return EventSourceResponse(event_generator())
-    except UpstreamServiceError:
-        request_id = get_request_id()
-        logger.exception(f"Upstream service error for request {request_id}")
-        return EventSourceResponse(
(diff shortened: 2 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -55,6 +55,7 @@ repos:
     hooks:
       - id: mypy
         name: mypy-clients-python
+        files: clients/python/.*
```
yunfeng-scale, line 58: shouldn't be using clients/python mypy checking server files

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -226,42 +226,38 @@ async def create_completion_stream_task(
...
-            llm_model_endpoint_service=external_interfaces.llm_model_endpoint_service,
-        )
-        response = use_case.execute(
-            user=auth, model_endpoint_name=model_endpoint_name, request=request
-        )
+    use_case = CompletionStreamV1UseCase(
+        model_endpoint_service=external_interfaces.model_endpoint_service,
+        llm_model_endpoint_service=external_interfaces.llm_model_endpoint_service,
+    )
+    response = use_case.execute(user=auth, model_endpoint_name=model_endpoint_name, request=request)
```
yunfeng-scale, line 233: here `use_case.execute` doesn't actually execute until L237

model-engine/tests/unit/api/test_llms.py
```diff
@@ -113,6 +113,32 @@ def test_completion_sync_success(
...
+    )
+    response_1 = client.post(
+        f"/v1/llm/completions-sync?model_endpoint_name={llm_model_endpoint_sync[0].record.name}",
+        auth=("no_user", ""),
+        json=completion_sync_request,
+    )
+    assert response_1.status_code == 404
+
+
 @pytest.mark.skip(reason="Need to figure out FastAPI test client asyncio funkiness")
```
saiatmakuri, line 142: do we still want to skip this test?

model-engine/tests/unit/api/test_llms.py
```diff
@@ -146,3 +173,34 @@ def test_completion_stream_success(
...
+        f"/v1/llm/completions-stream?model_endpoint_name={llm_model_endpoint_streaming.record.name}",
+        auth=("no_user", ""),
+        json=completion_stream_request,
+        stream=True,
+    )
+
+    assert response_1.status_code == 200
+
+    for message in response_1:
+        print(message)
```
saiatmakuri, line 205: nit: remove

saiatmakuri (approved):

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -226,42 +226,37 @@ async def create_completion_stream_task(
...
+        except InvalidRequestException as exc:
+            yield {"data": {"error": {"status_code": 400, "detail": str(exc)}}}
+        except UpstreamServiceError as exc:
+            request_id = get_request_id()
+            logger.exception(f"Upstream service error for request {request_id}")
+            yield {"data": {"error": {"status_code": 500, "detail": str(exc)}}}
+        except (ObjectNotFoundException, ObjectNotAuthorizedException) as exc:
+            yield {"data": {"error": {"status_code": 404, "detail": str(exc)}}}
+        except ObjectHasInvalidValueException as exc:
+            yield {"data": {"error": {"status_code": 400, "detail": str(exc)}}}
```
yixu34, line 248: Should we have a helper function to generate these payloads?

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -226,42 +226,37 @@ async def create_completion_stream_task(
...
+        except EndpointUnsupportedInferenceTypeException as exc:
```
yixu34, line 249: Should there be a catch-all that returns a 500?

model-engine/model_engine_server/api/llms_v1.py
yunfeng-scale: i think this snippet is short enough so okay to not have a helper function

model-engine/model_engine_server/api/llms_v1.py
yunfeng-scale: surely

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -71,6 +75,37 @@
...
+    structured_log = {
+        "error": message,
+        "request_id": str(request_id),
+        "traceback": "".join(tb_str),
+    }
+    logger.error("Exception: %s", structured_log)
+    return {
+        "data": json.dumps(
+            {
+                "request_id": str(request_id),
```
yixu34, line 95: `request_id` appears twice? Also this one is being `str`-ed, but the other one isn't?

yixu34 (commented):
Should also have @song-william take a look

model-engine/model_engine_server/api/llms_v1.py
yunfeng-scale: err, good point. removing the other one

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -71,6 +75,36 @@
...
+                "error": {
+                    "status_code": code,
```
yixu34, line 97: Maybe this is a bit pedantic (pydantic? lol), but since you ended up creating the DTOS anyway, wonder if it makes sense to instantiate the DTOs and convert them to JSON? That way you get the typing.

Alternatively, I think we want a unit test to enforce the API contract behind the error return type.

yunfeng-scale: will do. for test i added `test_completion_stream_endpoint_not_found_returns_404` but unfortunately i still haven't figured out how to run it

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -226,42 +259,30 @@ async def create_completion_stream_task(
...
-                    yield {"data": message.json()}
-            except InvalidRequestException as exc:
-                yield {"data": {"error": {"status_code": 400, "detail": str(exc)}}}
-                return
+    async def event_generator():
+        try:
+            async for message in response:
+                yield {"data": message.json()}
+        except (InvalidRequestException, ObjectHasInvalidValueException) as exc:
+            yield handle_streaming_exception(exc, 400, str(exc))
```
song-william, line 273: Just for my edification, was the primary fix for `urllib3.exceptions.ProtocolError` to push the exception handling inside the generator with `yield`?

song-william (approved):

model-engine/model_engine_server/api/llms_v1.py
yunfeng-scale: exception in `use_case.execute` won't throw until `async for message in response:`, and since originally we don't capture exceptions other than `InvalidRequestException`, my understanding is 0 bytes would get returned and client side throws error about not able to decode

PR #317, opened 2023-10-10
Add AWQ enum

yunfeng-scale:


ian-scale:
do we also need to add to spellbook-backend? or egp apis? I think both may also be affected

ian-scale:
> do we also need to add to spellbook-backend? or egp apis? I think both may also be affected

just checked and it doesn't seem like Quantization is present there, looks good!

ian-scale (approved):

yunfeng-scale:
> > do we also need to add to spellbook-backend? or egp apis? I think both may also be affected
> 
> just checked and it doesn't seem like Quantization is present there, looks good!

yeah we don't send back these details

PR #318, opened 2023-10-11
Properly return PENDING status for docker image batch jobs/fine tune jobs

seanshi-scale:
Before, jobs would immediately start as RUNNING even if no pods were allocated, or if pods were PENDING.

testing:
started gateway on devbox, made requests to get/list docker image batch jobs on our clusters, saw that a RUNNING job had changed to PENDING correctly.
also unit tests

### DIFF
--- a/model-engine/model_engine_server/infra/gateways/live_cron_job_gateway.py
+++ b/model-engine/model_engine_server/infra/gateways/live_cron_job_gateway.py
@@ -111,7 +111,7 @@ class LiveCronJobGateway(CronJobGateway):
             logger.exception("Got an exception when trying to list the Pods")
             raise EndpointResourceInfraException from exc
 
-        pods_per_job = make_job_id_to_pods_mapping(pods)
+        pods_per_job = make_job_id_to_pods_mapping(pods.items)
 
         return [
             DockerImageBatchJob(
--- a/model-engine/model_engine_server/infra/gateways/live_cron_job_gateway.py
+++ b/model-engine/model_engine_server/infra/gateways/live_cron_job_gateway.py
@@ -10,9 +10,11 @@ from model_engine_server.domain.gateways.cron_job_gateway import CronJobGateway
 from model_engine_server.infra.gateways.live_docker_image_batch_job_gateway import (
     LAUNCH_JOB_ID_LABEL_SELECTOR,
     _parse_job_status_from_k8s_obj,
+    make_job_id_to_pods_mapping,
 )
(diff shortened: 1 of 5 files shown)


model-engine/model_engine_server/infra/gateways/live_cron_job_gateway.py
```diff
@@ -97,14 +99,30 @@ async def list_jobs(
             logger.exception("Got an exception when trying to list the Jobs")
             raise EndpointResourceInfraException from exc
 
+        core_client = get_kubernetes_core_client()
```
seanshi-scale, line 102: apparently this fn is used in the `list docker image batch jobs` api call for some reason, feels wrong to me, idk how we didn't catch this before, but oh well

yixu34 (commented):
Can we add some unit tests for these gateways? We could mock out the k8s layer, since we're at the end of the line anyway.

model-engine/model_engine_server/infra/gateways/live_docker_image_batch_job_gateway.py
```diff
@@ -94,10 +97,27 @@ def _parse_job_status_from_k8s_obj(job: V1Job) -> BatchJobStatus:
     if status.ready is not None and status.ready > 0:
         return BatchJobStatus.RUNNING  # empirically this doesn't happen
     if status.active is not None and status.active > 0:
-        return BatchJobStatus.RUNNING  # TODO this might be a mix of pending and running
+        for pod in pods:
```
song-william, line 100: Might be worth leaving a comment here on why a single Job resource could have multiple pods due to clarify the logic here.

model-engine/model_engine_server/infra/gateways/live_cron_job_gateway.py
song-william: Could you like to the code that is the issue? Is the problem that we are breaking abstraction layers?

song-william (approved):

model-engine/model_engine_server/infra/gateways/live_cron_job_gateway.py
seanshi-scale: `model_engine_server/domain/use_cases/batch_job_use_cases.py:ListDockerImageBatchJobV1UseCase.execute`

This feels like we're breaking abstraction layers to me at least (e.g. batch jobs and cron jobs should be different), although I guess that broken abstraction gets propagated through the API as well (since the ListBatchJobs has a trigger_id parameter)

PR #327, opened 2023-10-16
create celery worker with inference worker profile

saiatmakuri:
# Pull Request Summary
The endpoint-builder and celery forwarder both use this celery app sessions instantiations. However, they expect different profiles to be used. Force the celery worker to use the dynamic value in `AWS_PROFILE` instead of the static `profile_ml_worker`.

## Testing Plan
Create a new deployment in our clusters and launched a batch job. The pods were built correctly and the job succeeded.

### DIFF
--- a/model-engine/model_engine_server/core/celery/app.py
+++ b/model-engine/model_engine_server/core/celery/app.py
@@ -504,7 +504,7 @@ def _get_backend_url_and_conf(
     elif backend_protocol == "s3":
         backend_url = "s3://"
         if aws_role is None:
-            aws_session = session(infra_config().profile_ml_worker)
+            aws_session = session(infra_config().profile_ml_inference_worker)
         else:
             aws_session = session(aws_role)
         out_conf_changes.update(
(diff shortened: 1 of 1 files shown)


yunfeng-scale:
can you briefly describe what this fixes?

model-engine/model_engine_server/core/celery/app.py
```diff
@@ -504,7 +504,7 @@ def _get_backend_url_and_conf(
     elif backend_protocol == "s3":
         backend_url = "s3://"
         if aws_role is None:
-            aws_session = session(infra_config().profile_ml_worker)
+            aws_session = session(infra_config().profile_ml_inference_worker)
```
yixu34, line 507: Are we sure this doesn't break things?

seanshi-scale: actually wait this probably breaks the endpoint builder on launch (endpoint builder on launch probably uses ml-worker), I think we need to look at `AWS_PROFILE` and use that value (something along those lines anyways)

seanshi-scale (approved):
lgtm as long as we've tested endpoint builder/inference workers, which I think creating a batch job will end up hitting both

PR #320, opened 2023-10-12
[Docs] Clarify get-events API usage

seanshi-scale:
Also update docs to reflect that mistral-7b is available

PR #332, opened 2023-10-18
Add docs for Model.create, update default values and fix per_worker concurrency

yunfeng-scale:
image
image

### DIFF
--- a/clients/python/llmengine/model.py
+++ b/clients/python/llmengine/model.py
@@ -167,6 +167,94 @@ class Model(APIEngine):
                 An optional dictionary of key/value pairs to associate with this endpoint
         Returns:
             CreateLLMEndpointResponse: creation task ID of the created Model. Currently not used.
+
+        === "Create Llama 2 7B model in Python"
+            ```python
+            from llmengine import Model
+
+            response = Model.create(
+                name="llama-2-7b-test"
+                model="llama-2-7b",
+                inference_framework_image_tag="2.0.0",
+                inference_framework=LLMInferenceFramework.VLLM,
+                num_shards=1,
+                checkpoint_path="s3://path/to/checkpoint",
+                cpus=8,
+                memory="24Gi",
+                storage="40Gi",
+                gpus=1,
+                min_workers=0,
+                max_workers=1,
+                per_worker=10,
+                endpoint_type=ModelEndpointType.STREAMING,
+                gpu_type="nvidia-ampere-a10",
+                public_inference=False,
+                labels={
+                    "product": "llm_model_zoo"
+                }
+            )
+
+            print(response.json())
+            ```
+
+        === "Create Llama 2 13B model in Python"
+            ```python
+            from llmengine import Model
+
+            response = Model.create(
+                name="llama-2-13b-test"
+                model="llama-2-13b",
+                inference_framework_image_tag="2.0.0",
+                inference_framework=LLMInferenceFramework.VLLM,
+                num_shards=2,
+                checkpoint_path="s3://path/to/checkpoint",
+                cpus=16,
+                memory="48Gi",
+                storage="80Gi",
+                gpus=2,
+                min_workers=0,
+                max_workers=1,
+                per_worker=10,
+                endpoint_type=ModelEndpointType.STREAMING,
+                gpu_type="nvidia-ampere-a10",
+                public_inference=False,
+                labels={
+                    "product": "llm_model_zoo"
+                }
+            )
+
+            print(response.json())
+            ```
+
+        === "Create Llama 70B model with 8bit quantization in Python"
+            ```python
+            from llmengine import Model
+
+            response = Model.create(
+                name="llama-2-70b-test"
+                model="llama-2-70b",
+                inference_framework_image_tag="0.9.4",
+                inference_framework=LLMInferenceFramework.TEXT_GENERATION_INFERENCE,
+                num_shards=4,
+                quantize="bitsandbytes",
+                checkpoint_path="s3://path/to/checkpoint",
+                cpus=40,
+                memory="96Gi",
+                storage="200Gi",
+                gpus=4,
+                min_workers=0,
+                max_workers=1,
+                per_worker=10,
+                endpoint_type=ModelEndpointType.STREAMING,
+                gpu_type="nvidia-ampere-a10",
+                public_inference=False,
+                labels={
+                    "product": "llm_model_zoo"
+                }
+            )
+
+            print(response.json())
+            ```
         """
         post_inference_hooks_strs = None
         if post_inference_hooks is not None:
--- a/clients/python/llmengine/model.py
+++ b/clients/python/llmengine/model.py
@@ -42,8 +42,8 @@ class Model(APIEngine):
         checkpoint_path: Optional[str] = None,
         # General endpoint fields
         cpus: int = 8,
-        memory: str = "20Gi",
-        storage: str = "96Gi",
+        memory: str = "24Gi",
+        storage: str = "40Gi",
         gpus: int = 1,
         min_workers: int = 0,
         max_workers: int = 1,
--- a/clients/python/llmengine/model.py
+++ b/clients/python/llmengine/model.py
@@ -73,18 +73,19 @@ class Model(APIEngine):
                 Source of the LLM. Currently only HuggingFace is supported
 
             inference_framework (`LLMInferenceFramework`):
-                Inference framework for the LLM. Currently supported frameworks are DeepSpeed, text-generation-inference, vLLM, and LightLLM
+                Inference framework for the LLM. Currently supported frameworks are `deepspeed`, `text_generation_inference`, `vllm`, and `lightllm`
 
             num_shards (`int`):
                 Number of shards for the LLM. When bigger than 1, LLM will be sharded
                 to multiple GPUs. Number of GPUs must be equal or larger than num_shards.
 
             quantize (`Optional[Quantization]`):
-                Quantization method for the LLM. Different frameworks support different set of quantization methods.
+                Quantization method for the LLM. `text_generation_inference` supports `bitsandbytes` and `vllm` supports `awq`.
 
             checkpoint_path (`Optional[str]`):
-                Path to the checkpoint for the LLM. Can be either a folder (preferred since there's no untar) or a tar file.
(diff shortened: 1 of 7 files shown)


clients/python/llmengine/model.py
```diff
@@ -72,41 +73,41 @@ def create(
                 Source of the LLM. Currently only HuggingFace is supported
 
             inference_framework (`LLMInferenceFramework`):
-                Inference framework for the LLM. Currently only DeepSpeed is supported
+                Inference framework for the LLM. Currently supported frameworks are DeepSpeed, text-generation-inference, vLLM, and LightLLM
```
yixu34, line 76: We should reference the literal `LLMInferenceFramework` values.

clients/python/llmengine/model.py
```diff
@@ -72,41 +73,41 @@ def create(
...
 
             num_shards (`int`):
                 Number of shards for the LLM. When bigger than 1, LLM will be sharded
                 to multiple GPUs. Number of GPUs must be equal or larger than num_shards.
-                Only affects behavior for text-generation-inference models
 
             quantize (`Optional[Quantization]`):
-                Quantization for the LLM. Only affects behavior for text-generation-inference models
+                Quantization method for the LLM. Different frameworks support different set of quantization methods.
```
yixu34, line 83: Think it's worth writing the table here.

clients/python/llmengine/model.py
```diff
@@ -72,41 +73,41 @@ def create(
...
 
             checkpoint_path (`Optional[str]`):
-                Path to the checkpoint for the LLM. For now we only support loading a tar file from AWS S3.
+                Path to the checkpoint for the LLM. Can be either a folder (preferred since there's no untar) or a tar file.
```
yixu34, line 86: > Can be either a folder

I think we want to be precise and consistent around our guidance relating to remote files/directories. Would suggest cross-checking with the Files API and making sure that the explanation around this makes sense. e.g. we may want to clarify that this is meant to be some remote path that's accessible by the LLM Engine deployment, etc.

> Can be either a folder (preferred since there's no untar)
Might be good to explain why skipping the untar is desirable - is it cold start time?

clients/python/llmengine/model.py
```diff
@@ -115,33 +116,19 @@ def create(
...
-                ``per_worker``, then the number of workers will be increased to meet the elevated
-                traffic.
-
-                Here is our recommendation for computing ``per_worker``:
-
-                1. Compute ``min_workers`` and ``max_workers`` per your minimum and maximum
-                throughput requirements. 2. Determine a value for the maximum number of
-                concurrent requests in the workload. Divide this number by ``max_workers``. Doing
-                this ensures that the number of workers will "climb" to ``max_workers``.
+                Number of Uvicorn workers per pod. Recommendation is set to 2.
```
yixu34, line 119: Might want to hide details of Uvicorn.

Also I don't think this is correct? IIRC `per_worker` is basically a scaling sensitivity parameter. cc @seanshi-scale

clients/python/llmengine/model.py
```diff
@@ -115,33 +116,19 @@ def create(
...
 
             endpoint_type (`ModelEndpointType`):
-                ``"sync"``, ``"async"`` or ``"streaming"``.
+                ``"sync"``, ``"async"`` or ``"streaming"``. Recommendation is ``"streaming"``.
```
yixu34, line 122: Might want to give some context as to what the endpoint types and what the tradeoffs are, basically giving more color to why the recommendation is streaming.

Also I thought async is not supported in LLM Engine? Should we take that out?

clients/python/llmengine/model.py
```diff
@@ -115,33 +116,19 @@ def create(
...
 
             gpu_type (`Optional[str]`):
                 If specifying a non-zero number of gpus, this controls the type of gpu
-                requested. Here are the supported values:
+                requested. Here are some supported values:
```
yixu34, line 126: I think we should leave it as `the` and be complete.

clients/python/llmengine/model.py
yunfeng-scale: checking code, i don't think we only support creating endpoints from files created with Files API.

yes skipping untar is good to cold start time

clients/python/llmengine/model.py
yunfeng-scale: hmm, this actually decides both # of workers and HPA target concurrency https://github.com/scaleapi/llm-engine/blob/main/model-engine/model_engine_server/infra/gateways/k8s_resource_parser.py#L65
i'll make some updates

clients/python/llmengine/model.py
yunfeng-scale: actually we do support async endpoint for deepspeed. i think we should just say we only support streaming

clients/python/llmengine/model.py
```diff
@@ -72,41 +73,43 @@ def create(
                 Source of the LLM. Currently only HuggingFace is supported
 
             inference_framework (`LLMInferenceFramework`):
-                Inference framework for the LLM. Currently only DeepSpeed is supported
+                Inference framework for the LLM. Currently supported frameworks are `deepspeed`, `text_generation_inference`, `vllm`, and `lightllm`
```
yixu34, line 76: How would a user make use of `vllm`? Would they pass in `inference_framework=vllm`, or `inference_framework="vllm"`...? It seems like we need to be precise about the literal value, e.g. `LLMInferenceFramework.VLLM`.

clients/python/llmengine/model.py
```diff
@@ -72,41 +73,43 @@ def create(
...
 
             quantize (`Optional[Quantization]`):
-                Quantization for the LLM. Only affects behavior for text-generation-inference models
+                Quantization method for the LLM. `text_generation_inference` supports `bitsandbytes` and `vllm` supports `awq`.
 
             checkpoint_path (`Optional[str]`):
-                Path to the checkpoint for the LLM. For now we only support loading a tar file from AWS S3.
-                Safetensors are preferred but PyTorch checkpoints are also accepted (model loading will be slower).
-                Only affects behavior for text-generation-inference models
+                AWS S3 Path to the checkpoint for the LLM. LLM engine must have permission to access the given path.
```
yixu34, line 86: I think we can say "Remote path" instead of "AWS S3 Path", since this should technically be OK as long as LLM Engine can access the file. And restricting it to S3 is not what we want for self-hosted users.

clients/python/llmengine/model.py
```diff
@@ -72,41 +73,43 @@ def create(
...
 
             gpus (`int`):
                 Number of gpus each worker should get, e.g. 0, 1, etc.
 
             min_workers (`int`):
                 The minimum number of workers. Must be greater than or equal to 0. This
                 should be determined by computing the minimum throughput of your workload and
-                dividing it by the throughput of a single worker. This field must be at least ``1``
-                for synchronous endpoints
+                dividing it by the throughput of a single worker. For sync or streaming endpoints,
```
yixu34, line 109: https://github.com/scaleapi/llm-engine/pull/332/files#diff-3147c6069e71e6e228995eebcaeee13595a6d2bbe8b389679c3f2905baf8090bR137

If we're saying that only `streaming` endpoints are supported, should we remove the `For sync or streaming endpoints, " phrase then?

model-engine/model_engine_server/infra/gateways/k8s_resource_parser.py
```diff
@@ -2,9 +2,6 @@
 import re
 from typing import Union
 
-MAX_CONCURRENCY_TO_TARGET_CONCURRENCY_RATIO = 2.0
```
yixu34, line 5: Wouldn't this technically change our autoscaling sensitivity? cc @seanshi-scale

yixu34: Which might be fine, just being aware of production behavior changes.

yunfeng-scale: yes this would but i think people are not aware of this ratio at all (like users specify per worker concurrency target to be 10 and HPA uses 5)

clients/python/llmengine/model.py
```diff
@@ -57,7 +57,8 @@ def create(
         labels: Optional[Dict[str, str]] = None,
     ) -> CreateLLMEndpointResponse:
         """
-        Create an LLM model. Note: This feature is only available for self-hosted users.
+        Create an LLM model. Note: This API is only available for self-hosted users.
```
yixu34, line 60: One more high-level comment: can we add some example code snippets of how a user would invoke `Model.create()` for a few different models in our Model Zoo? e.g. concrete examples for Llama-2-7B, 13B, 70B, etc. This is more to provide guidance for internal users, as a stopgap until we do the ticket about removing all these knobs.

Made a similar comment here, about adding examples: https://github.com/scaleapi/llm-engine/pull/320#discussion_r1358664717

clients/python/llmengine/model.py
```diff
@@ -167,6 +167,94 @@ def create(
...
             CreateLLMEndpointResponse: creation task ID of the created Model. Currently not used.
+
+        === "Create Llama 2 7B model in Python"
+            ```python
+            from llmengine import Model
+
+            response = Model.create(
+                name="llama-2-7b-test"
+                model="llama-2-7b",
+                inference_framework_image_tag="2.0.0",
```
yixu34, line 178: Is this a valid inference framework tag for vLLM? Thought this should be `"0.2.0"`? Though it looks like we may want to use `"0.2.1"` for a recent fix.

clients/python/llmengine/model.py
```diff
@@ -167,6 +167,94 @@ def create(
...
+                memory="24Gi",
+                storage="40Gi",
+                gpus=1,
+                min_workers=0,
+                max_workers=1,
+                per_worker=10,
+                endpoint_type=ModelEndpointType.STREAMING,
+                gpu_type="nvidia-ampere-a10",
+                public_inference=False,
+                labels={
```
yixu34, line 192: We might want to omit this, or put a separate `"product": "undefined"` value. We only want this particular `"llm_model_zoo"` label when we create the base model endpoints ourselves - we don't want other teams to use this.

yixu34 (commented):
Only thing I had left were around the examples:
1. framework version tag
2. labels

yixu34 (approved):

PR #334, opened 2023-10-19
Hardcode number of forwarder workers

yunfeng-scale:
hardcode to use 2 http forwarder workers:
1. we're only assigning 500m CPU to forwarder
2. given our traffic don't think we need more than 2


yixu34 (approved):

charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -217,7 +217,7 @@ data:
                 - "${SQS_QUEUE_URL}"
                 {{- end }}
                 - --num-workers
-                - "${PER_WORKER}"
+                - "${FORWARDER_WORKER_COUNT}"
```
yunfeng-scale: we always just have one celery worker and `per_worker` is used to determine the concurrency https://github.com/scaleapi/llm-engine/blob/main/model-engine/model_engine_server/inference/forwarding/celery_forwarder.py#L141

seanshi-scale: imo we could think of ways to maybe increase the number of celery tasks in flight to increase async task throughput, which may involve increasing number of workers, maybe not super necessary now though

PR #335, opened 2023-10-19
Fix up the mammoth max length issue.

sam-scale:


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -514,13 +534,14 @@ async def create_vllm_bundle(
...
         command = []
 
-        max_num_batched_tokens = 2560  # vLLM's default
-        max_model_len = None
-        if "llama-2" in model_name:
-            max_num_batched_tokens = 4096  # Need to be bigger than model's context window
-        if "mistral" in model_name:
-            max_num_batched_tokens = 8000
-            max_model_len = 8000
+        max_num_batched_tokens: Optional[int] = 2560  # vLLM's default
```
yunfeng-scale, line 537: When is max_num_batched_tokens None?

sam-scale: Never, it's because of the typing of the override dict `_VLLM_MODEL_LENGTH_OVERRIDES: Dict[str, Dict[str, Optional[int]]] `

I can try a better solution

yunfeng-scale (approved):

PR #338, opened 2023-10-20
ensuring invalid image tag errors are surfaced to users clearly

ian-scale:
This change will now enforce that users specify valid `inference_framework` / `inference_framework_image_tag` pairs when trying to invoke Model.create(). I think we should add this because users have been passing in invalid image tags and then getting confused about why their creation fails. However, it's worth noting that we will have to remember to update this dictionary whenever we add a new version of one of our inference frameworks.


yixu34:
Closing in favor of #357

PR #341, opened 2023-10-20
Add PR template

song-william:
Add a template for Pull Requests in this repo


yixu34 (approved):
Would suggest look at PR templates from other OSS projects. Think we'd want a form field for the associated GH issue, if applicable.

PR #345, opened 2023-10-23
Revert "fix celery worker profile for s3 access"

yixu34:
Reverts scaleapi/llm-engine#333

I think this caused an issue:
```
[2023-10-23 20:00:55,721: ERROR/MainProcess] Pool callback raised exception: ClientError('An error occurred (AccessDenied) when calling the PutObject operation: Access Denied')
```


yixu34:
Going to force merge this as it's breaking people's workflows, plus it's a revert.

PR #352, opened 2023-10-25
adding real auth to integration tests

ian-scale:


yixu34 (approved):

yunfeng-scale:
we should remove `SKIP_AUTH` in justfile/_integration_test?

ian-scale:
> we should remove `SKIP_AUTH` in justfile/_integration_test?

yes once the other changes I'm adding work, but still waiting to make sure integration tests pass locally

yunfeng-scale:
> > we should remove `SKIP_AUTH` in justfile/_integration_test?
> 
> yes once the other changes I'm adding work, but still waiting to make sure integration tests pass locally

are we skipping auth for circleci integration tests?

integration_tests/test_fine_tunes.py
```diff
@@ -1,7 +1,10 @@
+"""
```
ian-scale, line 1: this test was already not functioning so I commented everything out - todo later is come through and fix it so that it works.

yixu34: Mind creating a follow-up ticket for this?

yixu34 (approved):

integration_tests/test_fine_tunes.py
ian-scale: yeah sounds good!

integration_tests/rest_api_utils.py
```diff
@@ -15,17 +15,10 @@
...
-USER_PREFIX = os.getenv("SERVICE_IDENTIFIER", "test")[:8]
-USER_ID_0 = USER_PREFIX + "0" * (24 - len(USER_PREFIX))
-USER_ID_1 = USER_PREFIX + "1" * (24 - len(USER_PREFIX))
-
-DEFAULT_USERS: Sequence[str] = (
-    USER_ID_0,
-    USER_ID_1,
-)
+# Use the scale-launch-integration-tests id
+USER_ID_0 = os.getenv("TEST_USER_ID")  # type: ignore
```
yunfeng-scale, line 19: set some default value

integration_tests/test_fine_tunes.py
```diff
@@ -21,11 +24,20 @@ def test_fine_tunes() -> None:
     # num_jobs = len(list_response_0_before["jobs"])
     # assert num_jobs >= 1
 
-    list_response_1 = list_fine_tunes(USER_ID_1)
-    assert len(list_response_1["jobs"]) == 0
-
-    # cancel_response = cancel_fine_tune_by_id(fine_tune_id, USER_ID_0)
-    # assert cancel_response["success"]
+    # list_response_1 = list_fine_tunes(USER_ID_0)
```
seanshi-scale, line 27: there should be some `@pytest.skip(reason="test doesn't work")` thing that you can do to skip the test (as opposed to commenting it out)

ian-scale: yeah let me add this too - although I will say the test was already doing nothing

integration_tests/rest_api_utils.py
yunfeng-scale: and i think `None` USER_ID_0 caused integration test to fail

ian-scale: agreed, I set a default value of "fakeuser"

PR #356, opened 2023-10-30
Track LLM Metrics

saiatmakuri:
Emit metrics on each route calls for the LLM endpoints.

Tested locally with:
```bash
curl -H "Authorization: <AUTH>" http://localhost:5001/v1/llm/model-endpoints
```

```bash
curl -H "Content-Type: application/json" -H "Authorization: <AUTH>" -d '{ "prompt": "Hello!", "max_new_tokens": 10, "temperature": 0.1 }' http://localhost:5001/v1/llm/completions-sync?model_endpoint_name=llama-2-7b-test-vllm
```

### DIFF
--- a/model-engine/model_engine_server/api/llms_v1.py
+++ b/model-engine/model_engine_server/api/llms_v1.py
@@ -88,7 +88,7 @@ async def record_route_call(
     model_name = request.query_params.get("model_endpoint_name", None)
 
     external_interfaces.monitoring_metrics_gateway.emit_route_call_metric(
-        route=route, metadata=MetricMetadata(user=auth, model_name=model_name)
+        route, MetricMetadata(user=auth, model_name=model_name)
     )
 
 
--- a/model-engine/model_engine_server/api/llms_v1.py
+++ b/model-engine/model_engine_server/api/llms_v1.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from typing import Optional
 
 import pytz
-from fastapi import APIRouter, Depends, HTTPException, Query
+from fastapi import APIRouter, Depends, HTTPException, Query, Request
 from model_engine_server.api.dependencies import (
     ExternalInterfaces,
     get_external_interfaces,
@@ -78,7 +78,21 @@ from model_engine_server.domain.use_cases.llm_model_endpoint_use_cases import (
 from model_engine_server.domain.use_cases.model_bundle_use_cases import CreateModelBundleV2UseCase
 from sse_starlette.sse import EventSourceResponse
 
-llm_router_v1 = APIRouter(prefix="/v1/llm")
+
+async def record_route_call(
+    request: Request,
+    auth: User = Depends(verify_authentication),
+    external_interfaces: ExternalInterfaces = Depends(get_external_interfaces_read_only),
+):
+    route = f"{request.method}_{request.url.path}".lower()
+    model_name = request.query_params.get("model_endpoint_name", None)
+
+    external_interfaces.monitoring_metrics_gateway.emit_route_call_metric(
+        route=route, metadata=MetricMetadata(user=auth, model_name=model_name)
+    )
+
+
+llm_router_v1 = APIRouter(prefix="/v1/llm", dependencies=[Depends(record_route_call)])
 logger = make_logger(logger_name())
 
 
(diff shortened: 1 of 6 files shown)


model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -118,6 +119,9 @@ async def create_model_endpoint(
     """
     Creates an LLM endpoint for the current user.
     """
+    external_interfaces.monitoring_metrics_gateway.emit_route_call_metric(
```
yixu34, line 122: Should we emit a metric here? Or add tags to the current trace? cc @song-william

model-engine/model_engine_server/infra/gateways/fake_monitoring_metrics_gateway.py
```diff
@@ -57,3 +62,6 @@ def emit_database_cache_hit_metric(self):
 
     def emit_database_cache_miss_metric(self):
         self.database_cache_miss += 1
+
+    def emit_route_call_metric(self, route: str, _metadata: MetricMetadata):
```
yixu34, line 66: Maybe outside of the scope of this PR, but might be good to adopt Datadog-esque terminology, where `increment` means += 1. `emit` to me sounds more like a gauge or count.

model-engine/model_engine_server/api/llms_v1.py
song-william: I think adding tags to the current trace would "kill 2 birds with one stone" so we can search for traces based on user_id as well. 

Metrics generated by Traces are not sampled out: https://docs.datadoghq.com/tracing/metrics/metrics_namespace/

song-william: @saiatmakuri If we add a tag to the existing trace, then add this logic into the `auth` dependency. This would then make it so we don't need to copy-pasta this call to each route.

song-william: Ahh, it looks like we can't customize the tags for trace-based metrics. 

> Trace metrics tags, possible tags are: env, service, version, resource, sublayer_type, sublayer_service, http.status_code, http.status_class, Datadog Agent tags (including the host and second primary tag). Note: Tags set on spans do not count and will not be available as tags for your traces metrics.
https://docs.datadoghq.com/tracing/metrics/metrics_namespace/ (edited)

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -172,6 +176,9 @@ async def list_model_endpoints(
     """
     Lists the LLM model endpoints owned by the current owner, plus all public_inference LLMs.
     """
+    external_interfaces.monitoring_metrics_gateway.emit_route_call_metric(
```
song-william, line 179: Main comment is if there is a way to do "automatically" do this without having to copy-pasta on each route and to automatically used the defined route name. 

I have looked into something like this before, and I think we have to put it into a dependency. We can't do it in the middleware because we need the auth dep in order to get the user_id.

song-william (approved):
Approving to unblock

model-engine/model_engine_server/api/llms_v1.py
song-william: https://stackoverflow.com/a/66634433

If we want to add user auth info as a tag, I think our hand is forced to add these obs additions as a dependency (as opposed to middleware)

model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -77,7 +78,21 @@
 from model_engine_server.domain.use_cases.model_bundle_use_cases import CreateModelBundleV2UseCase
 from sse_starlette.sse import EventSourceResponse
 
-llm_router_v1 = APIRouter(prefix="/v1/llm")
+
+async def record_route_call(
```
song-william, line 82: We could consider adding this at the highest level so all model-engine routes get it (e.g bundles/endpoits).

model-engine/model_engine_server/infra/gateways/fake_monitoring_metrics_gateway.py
song-william: Do you have a link to the "Datadog-esque terminology" you are referring to? What is currently here looks like it is just pattern matching the `emit_*` pattern.

song-william (approved):
Thanks for addressing the comments!

model-engine/model_engine_server/infra/gateways/fake_monitoring_metrics_gateway.py
yixu34: Yeah was thinking of https://docs.datadoghq.com/metrics/custom_metrics/dogstatsd_metrics_submission/
image

But yeah acknowledge that this is just following the existing naming pattern (which is what I think could be tweaked).

PR #357, opened 2023-11-01
Validating inference framework image tags

tiffzhao5:
Added unit test to this: https://github.com/scaleapi/llm-engine/pull/338

Needed to rebase so created a separate PR. 

Querying ECR to check for existing images for each inference framework.


yunfeng-scale:
thanks for the contributions!

i'm worried about the overhead of adding these tags / redeploy model engine server whenever we're adding a new image. possible to query ECR and check existence of the image?

tiffzhao5:
> thanks for the contributions!
> 
> i'm worried about the overhead of adding these tags / redeploy model engine server whenever we're adding a new image. possible to query ECR and check existence of the image?

hm I see that's a good suggestion, I'll give it a try and see if it works!

tiffzhao5:
> > thanks for the contributions!
> > i'm worried about the overhead of adding these tags / redeploy model engine server whenever we're adding a new image. possible to query ECR and check existence of the image?
> 
> hm I see that's a good suggestion, I'll give it a try and see if it works!

@yunfeng-scale I implemented querying ECR -- what are your thoughts?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1045,10 +1059,7 @@ def model_output_to_completion_output(
             tokens = None
             if with_token_probs:
                 tokens = [
-                    TokenOutput(
-                        token=model_output["tokens"][index],
-                        log_prob=list(t.values())[0],
-                    )
+                    TokenOutput(token=model_output["tokens"][index], log_prob=list(t.values())[0])
```
ian-scale, line 1062: only comment here is - can you run the pre-commit hooks? I don't think these files should be changing. Or it could be possible that the old files didn't conform to pre-commit, would you mind checking?

tiffzhao5: hm I ran `black.toml` and it left my code changes unchanged so not really sure if I should run anything else?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -230,21 +232,8 @@ def validate_num_shards(
...
-        )
-
-
-def validate_quantization(
-    quantize: Optional[Quantization], inference_framework: LLMInferenceFramework
-) -> None:
-    if quantize is not None and quantize not in _SUPPORTED_QUANTIZATIONS[inference_framework]:
-        raise ObjectHasInvalidValueException(
-            f"Quantization {quantize} is not supported for inference framework {inference_framework}. Supported quantization types are {_SUPPORTED_QUANTIZATIONS[inference_framework]}."
-        )
```
yunfeng-scale, line 247: revert these removals?

yunfeng-scale:
it seems a some commits were reverted, can you bring them back?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -276,6 +267,14 @@ async def create_model_bundle(
     ) -> ModelBundle:
         if source == LLMSource.HUGGING_FACE:
             if framework == LLMInferenceFramework.DEEPSPEED:
+                if not self.docker_repository.image_exists(
```
yunfeng-scale, line 270: nit: i think we could extract a function

tiffzhao5: where would you want the function to reside? I feel like it'd be a small one so not sure if it should be directly in the `CreateLLMModelEndpointV1UseCase` class

yixu34: I think putting the helper in this class is fine. Has less to do with the size of the function, as to its scope and relevance - IMO better to keep helpers "local" as much as possible unless they're needed outside of the module.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1351,7 +1372,7 @@ async def execute(
         )
 
         if len(model_endpoints) == 0:
-            raise ObjectNotFoundException(f"Model endpoint {model_endpoint_name} not found.")
+            raise ObjectNotFoundException
```
yixu34, line 1375: Was this change intentional?

yunfeng-scale (approved):
please fix @yixu34 's comment

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
tiffzhao5: Oops thanks for the catch, not sure how that got deleted, will fix

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
tiffzhao5: Gotcha sounds good, will refactor!

PR #358, opened 2023-11-01
Integrate TensorRT-LLM

yunfeng-scale:
Integrating TensorRT-LLM with LLM engine. there are lots of caveats in this integration:

- Stop word not working for TRT-LLM Triton backend https://github.com/triton-inference-server/tensorrtllm_backend/issues/57
- Log prob not returned https://github.com/NVIDIA/TensorRT-LLM/issues/238
- Need to verify definitions of presence_penalty and repetition_penalty that TRT-LLM supports
- Need to send back only newly generated tokens https://github.com/NVIDIA/TensorRT-LLM/issues/227
- Can't tell end of stream https://github.com/NVIDIA/TensorRT-LLM/issues/240

fixes https://github.com/scaleapi/llm-engine/issues/363


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -695,6 +733,70 @@ async def create_lightllm_bundle(
...
+        command = []
+
+        subcommands = []
+        if checkpoint_path is not None:
+            if checkpoint_path.startswith("s3://"):
+                subcommands += self.load_model_files_sub_commands_trt_llm(
+                    checkpoint_path,
+                )
+            else:
+                raise ObjectHasInvalidValueException(
```
yixu34, line 753: Given that the `if` is explicitly checking if it's an s3 path, maybe this error can be more specific and tell the user that only s3 paths are supported? A more generic "Not able to load checkpoint path" error can still be useful as a stopgap catch-all for `load_model_files_sub_commands_trt_llm`, until we can get more precise there too.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -983,9 +1086,30 @@ def validate_and_update_completion_params(
...
+        LLMInferenceFramework.DEEPSPEED,
+        LLMInferenceFramework.TEXT_GENERATION_INFERENCE,
+        LLMInferenceFramework.VLLM,
+        LLMInferenceFramework.LIGHTLLM,
+    ]:
+        pass
+    else:
+        if request.return_token_log_probs:
+            raise ObjectHasInvalidValueException(
+                "return_token_log_probs are only supported in deepspeed, text-generation-inference, vllm, lightllm."
```
yixu34, line 1100: nit: `is` instead of `are`

yixu34 (approved):
Some minor comments.

Can we create a github issue for visibility and link it to this PR?

PR #359, opened 2023-11-01
Remove extra trace facet "launch.resource_name"

saiatmakuri:
https://docs.datadoghq.com/tracing/metrics/metrics_namespace/
> Trace metrics tags, possible tags are: env, service, version, resource, sublayer_type, sublayer_service, http.status_code, http.status_class, Datadog Agent tags (including the host and second primary tag). Note: Tags set on spans do not count and will not be available as tags for your traces metrics.

There already is a resource trace metric tag. `launch.resource_name` is also a facet that's unused, so it can be removed

### DIFF
--- a/model-engine/model_engine_server/api/batch_jobs_v1.py
+++ b/model-engine/model_engine_server/api/batch_jobs_v1.py
@@ -7,7 +7,6 @@ from model_engine_server.api.dependencies import (
     get_external_interfaces_read_only,
     verify_authentication,
 )
-from model_engine_server.common.datadog_utils import add_trace_resource_name
 from model_engine_server.common.dtos.batch_jobs import (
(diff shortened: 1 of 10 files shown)


ian-scale:
@saiatmakuri I don't think this is the case, but can you confirm this facet isn't being used for any of our datadog dashboards before we remove it?

ian-scale (approved):

saiatmakuri:
> @saiatmakuri I don't think this is the case, but can you confirm this facet isn't being used for any of our datadog dashboards before we remove it?

checked again to confirm, we only use the trace metric resource_name

song-william (approved):

yixu34 (approved):
🧹 👌🏻

PR #366, opened 2023-11-07
count prompt tokens

saiatmakuri:
# Pull Request Summary

For all completions, return the number of tokens in the prompt as `num_prompt_tokens`.

We get the count for prompt tokens in a waterfall method of fallbacks:
- get prompt tokens count from the inference framework
- load a tokenizer from HF weights
- load a tokenizer from weights stored in S3

## Test Plan and Usage Guide

Ran skipped integration test on local test. Need to figure out how to enable for CircleCI.
Compared token counts to https://huggingface.co/spaces/Xenova/the-tokenizer-playground


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -178,6 +266,82 @@
 DOWNSTREAM_REQUEST_TIMEOUT_SECONDS = 5 * 60  # 5 minutes
 
 
+# Hack to count prompt tokens
+tokenizer_cache: Dict[str, AutoTokenizer] = {}
```
seanshi-scale, line 270: nit: would we want to make this an lru cache or something? idk how big the tokenizers can get

saiatmakuri: ~3MB per model. good call to add lru cache

model-engine/model_engine_server/common/tokenizer_utils.py
```diff
@@ -0,0 +1,173 @@
+from collections import namedtuple
```
yixu34, line 1: Would like to avoid a "common" / "utils" file as that starts to erode the clean architecture setup we have.

model-engine/model_engine_server/common/tokenizer_utils.py
```diff
@@ -0,0 +1,173 @@
...
+    "gpt-j-6b": ModelInfo("EleutherAI/gpt-j-6b", None),
+    "gpt-j-6b-zh-en": ModelInfo("EleutherAI/gpt-j-6b", None),
+    "gpt4all-j": ModelInfo("nomic-ai/gpt4all-j", None),
+    "dolly-v2-12b": ModelInfo("databricks/dolly-v2-12b", None),
+    "stablelm-tuned-7b": ModelInfo("StabilityAI/stablelm-tuned-alpha-7b", None),
+    "vicuna-13b": ModelInfo("eachadea/vicuna-13b-1.1", None),
+}
+
+
+def get_models_s3_uri(s3_prefix: str, file: str) -> str:
```
yixu34, line 105: Let's hide these.

model-engine/tests/unit/infra/gateways/test_s3_llm_artifact_gateway.py
```diff
@@ -0,0 +1,81 @@
...
+    return gateway
+
+
+@pytest.fixture
+def fake_files():
+    return ["fake-prefix/fake1", "fake-prefix/fake2", "fake-prefix/fake3"]
+
+
+def mock_boto3_session(fake_files: List[str]):
+    mock_session = mock.Mock()
```
yixu34, line 21: `moto` could be an option here, though this mocking can be fine too since you're already at the last layer, where you're in the S3 artifact gateway (want to avoid mocking across multiple layers).

model-engine/model_engine_server/common/dtos/llms.py
```diff
@@ -80,7 +80,7 @@ class GetLLMModelEndpointV1Response(BaseModel):
     """
 
     name: str
-    model_name: Optional[str] = None
+    model_name: str
```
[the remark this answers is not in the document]
saiatmakuri: it looks like `model_name` is always passed in. would it break clients upstream?

integration_tests/test_endpoints.py
```diff
@@ -234,3 +239,10 @@ def test_sync_streaming_model_endpoint(capsys):
                 )
         finally:
             delete_model_endpoint(create_endpoint_request["name"], user)
+
+
+@pytest.mark.skip(reason="test doesn't currently work, needs to figure out s3 fallback")
```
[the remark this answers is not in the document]
saiatmakuri: for the integration tests before deploy, I'd like to check s3 as well. I had changed this skip to since the comment to a skipif to just skip for the circleci env

model-engine/model_engine_server/common/dtos/llms.py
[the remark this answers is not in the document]
seanshi-scale: also this is on the response right? so it's actually technically not a breaking change? (gonna have to regen clients but the older clients should still work)

model-engine/tests/unit/infra/gateways/test_s3_llm_artifact_gateway.py
saiatmakuri: ah moto seems cool, useful to know for the future. wanted to create a custom side effect so this leans itself more easily to that

model-engine/model_engine_server/common/tokenizer_utils.py
yixu34: Also discussing offline - IMO I think we should have a Repository for the tokenizers, and have the compute that uses the tokenizers inside the inference-related usecase.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1471,6 +1508,11 @@ async def execute(
                 args["parameters"]["do_sample"] = False
             if request.return_token_log_probs:
                 args["parameters"]["return_details"] = True
+            num_prompt_tokens = count_tokens(
```
yunfeng-scale, line 1511: i'm okay we do tokenization for less used frameworks, but for more important models can we move tokenization into the framework itself?

song-william: What about moving this fall-back tokenization into the forwarder? That would then have less overhead in the gateway, which also supports high QPS routes like `get/post tasks`.

model-engine/model_engine_server/common/dtos/llms.py
```diff
@@ -143,6 +143,7 @@ class TokenOutput(BaseModel):
 
 class CompletionOutput(BaseModel):
     text: str
+    num_prompt_tokens: int
```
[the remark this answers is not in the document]
yunfeng-scale: i don't remember exactly, but i think having them optional could reduce duplicated returned data, since `num_completion_tokens` can be inferred, and `num_prompt_tokens` only need to be returned once

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
seanshi-scale: yeah IMO the forwarder feels like the more "natural" place to put counting tokens, this way we'd only have to download one tokenizer in the forwarder, and we offload the computation to something that scales up more in proportion with load

this does mean that the forwarder is gonna have to know to carry out this token-counting logic exactly when it's forwarding to an LLM though, which will mean there are different "modes" for the forwarder (e.g. not-LLM, where it just passes requests through, and LLM, where it maybe does some specific processing and then passes requests through)

saiatmakuri: upstreaming all changes to the framework will be the goal, but this is a temporary stopgap

saiatmakuri: > this does mean that the forwarder is gonna have to know to carry out this token-counting logic exactly when it's forwarding to an LLM though, which will mean there are different "modes" for the forwarder (e.g. not-LLM, where it just passes requests through, and LLM, where it maybe does some specific processing and then passes requests through)

thought about this for emitting token metrics in gateway vs forwarder as well. think its worth a larger discussion after this PR

yunfeng-scale (approved):
what's the current ephemeral disk size for model engine pods? should we add more?

model-engine/model_engine_server/infra/repositories/live_tokenizer_repository.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def get_models_local_dir_path(model_name: str) -> str:
+    """
+    Get the local directory path for a given model.
+    """
+    return f"{TOKENIZER_TARGET_DIR}/{model_name}"
+
+
+class LiveTokenizerRepository(TokenizerRepository):
```
yunfeng-scale, line 103: where is this repository injected? can you update `_get_external_interfaces`?

saiatmakuri:
> what's the current ephemeral disk size for model engine pods? should we add more?

we've set it to 128Mi, this should be sufficient atm

yunfeng-scale:
> > what's the current ephemeral disk size for model engine pods? should we add more?
> 
> we've set it to 128Mi, this should be sufficient atm

unclear to me whether this is enough since i think previously model engine got out of disk due to usage of 100MB space?

model-engine/model_engine_server/common/dtos/llms.py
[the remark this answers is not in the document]
yixu34: Oh yeah nm.

model-engine/model_engine_server/infra/repositories/live_tokenizer_repository.py
```diff
@@ -0,0 +1,149 @@
...
+TOKENIZER_FILES_OPTIONAL = [
+    "tokenizer.model",
+]
+TOKENIZER_TARGET_DIR = "/opt/.cache/model_engine_server/tokenizers"
+
+
+ModelInfo = namedtuple("ModelInfo", ["hf_repo", "s3_repo"])
+
+
+def get_default_supported_models_info() -> Dict[str, ModelInfo]:
```
yixu34, line 31: Oh yeah so what I meant by hiding was to basically prefix these module-level functions with underscores, so that a user doesn't get tempted to try and use this out of context.

model-engine/model_engine_server/infra/repositories/live_tokenizer_repository.py
```diff
@@ -0,0 +1,149 @@
...
+    "tokenizer_config.json",
+    "special_tokens_map.json",
+]
+TOKENIZER_FILES_OPTIONAL = [
+    "tokenizer.model",
+]
+TOKENIZER_TARGET_DIR = "/opt/.cache/model_engine_server/tokenizers"
+
+
+ModelInfo = namedtuple("ModelInfo", ["hf_repo", "s3_repo"])
```
yixu34, line 28: Would a Union type be more appropriate, given that it seems that the s3 thing is always None it seems?

Also might want to hide this too.

model-engine/model_engine_server/infra/repositories/live_tokenizer_repository.py
```diff
@@ -0,0 +1,149 @@
...
+
+        model_location = None
+        try:
+            if not model_info.hf_repo:
+                raise RepositoryNotFoundError("No HF repo specified for model.")
+            list_repo_refs(model_info.hf_repo)  # check if model exists in Hugging Face Hub
+            model_location = model_info.hf_repo
+            # AutoTokenizer handles file downloads for HF repos
+        except RepositoryNotFoundError:
+            model_location = self._load_tokenizer_from_s3(model_name, model_info.s3_repo)
```
yixu34, line 143: What's the behavior if this fails? Seems like the API layer would catch and throw a 500?

yixu34: Oh nevermind I'm guess there's a catch-all `None` response based on the `if not model_location` call below.

model-engine/tests/unit/infra/gateways/test_s3_llm_artifact_gateway.py
```diff
@@ -0,0 +1,85 @@
...
+
+    mock_bucket.download_file.return_value = None
+    return mock_session
+
+
+@mock.patch(
+    "model_engine_server.infra.gateways.s3_llm_artifact_gateway.os.makedirs",
+    lambda *args, **kwargs: None,  # noqa
+)
+def test_s3_llm_artifact_gateway_download_folder(llm_artifact_gateway, fake_files):
```
yixu34, line 41: Thank you for bumping up our test coverage! 🤜🏻 🤛🏻

yixu34 (approved):
Looks good for a V0 of token counting. If needed later on, we can consider in-framework tokenization and/or doing tokenization in the forwarder.

We should carefully monitor error rates and token latency/throughput along this rollout

model-engine/model_engine_server/infra/repositories/live_tokenizer_repository.py
saiatmakuri: dependency injection passes in non none where applicable

PR #372, opened 2023-11-13
Fine-tuning e2e integration test

tiffzhao5:
# Pull Request Summary

Add e2e integration test for fine-tuning. Also changed the base docker image for loading integration tests onto minikube to `python3.8:slim` instead since it originally took 10 minutes -- now it only takes 4 min! Let me know if this breaks anything else.

## Test Plan and Usage Guide

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,10 @@ jobs:
           command: |
             pushd $HOME/project
             kubectl port-forward svc/model-engine 5001:80 &
-            GIT_TAG=$CIRCLE_SHA1 pytest integration_tests
+            export AWS_ACCESS_KEY_ID=$CIRCLECI_AWS_ACCESS_KEY
+            export AWS_SECRET_ACCESS_KEY=$CIRCLECI_AWS_SECRET_KEY
+            export GIT_TAG=$CIRCLE_SHA1
+            pytest integration_tests
 
 executors:
   ubuntu-large:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,6 +121,7 @@ jobs:
           command: |
             sudo apt-get update && sudo apt-get install -y expect
             pushd $HOME/project/.circleci/resources
+            kubectl create namespace model-engine
             kubectl apply -f redis-k8s.yaml
             kubectl apply -f postgres-k8s.yaml
             kubectl create secret generic model-engine-postgres-credentials --from-literal=database_url=postgresql://postgres:circle_test@postgres.default:5432/circle_test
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -142,7 +143,7 @@ jobs:
           name: Pre-load integration test images to minikube
           command: |
             docker build -f model-engine/model_engine_server/inference/pytorch_or_tf.base.Dockerfile \
-            --build-arg BASE_IMAGE=pytorch/pytorch:1.7.1-cuda11.0-cudnn8-runtime \
+            --build-arg BASE_IMAGE=python:3.8-slim \
```
yunfeng-scale, line 146: interesting, does it mean we could use this for torch images in docker-compose.yml?

.circleci/config.yml
```diff
@@ -121,15 +121,16 @@ jobs:
           command: |
             sudo apt-get update && sudo apt-get install -y expect
             pushd $HOME/project/.circleci/resources
+            kubectl create namespace model-engine
```
yunfeng-scale, line 124: is integration test model-engine actually deployed in this namespace?

yunfeng-scale (commented):
thanks a bunch for adding this important integration test!

yixu34 (approved):
Thanks for adding this!

.circleci/config.yml
tiffzhao5: hm I'm not sure actually

.circleci/config.yml
tiffzhao5: seems like from this https://github.com/scaleapi/llm-engine/blob/main/.circleci/config.yml#L164 that the helm chart is installed in the default namespace but from here https://github.com/scaleapi/llm-engine/blob/tiffany/fine-tune-e2e/charts/model-engine/values_circleci.yaml#L97 the endpoints are deployed in the model-engine namespace

yunfeng-scale: `helm install model-engine model-engine --values model-engine/values_circleci_subst.yaml --set tag=$CIRCLE_SHA1 --atomic --debug` does not specify which namespace to install the chart. it installs a chart defined in `./model-engine` folder, with name `model-engine`

endpoint and model-engine don't need to be in the same namespace

yunfeng-scale: can you try to remove these if this is actually not used?

tiffzhao5: actually the namespace `model-engine` is used for endpoints and batch jobs which was why I needed to create a new postgres secret there

yunfeng-scale: is this specific for finetune integration test? i don't think this blocks any other integration tests

tiffzhao5: the postgres secret is specific because we're actually checking if the batch job gets created in the fine-tune integration test. there's one batch job integration test but it doesn't actually check for a successful creation which is why we didn't need this secret before.

PR #374, opened 2023-11-14
Use Variable name for Telemetry Helm Resources

song-william:
# Pull Request Summary
Using variable name to remain consistent with other Helm resources.

## Test Plan and Usage Guide

Verified resource is valid and variables properly filled in.
```
helm template --debug model-engine model-engine -s templates/istio-metrics.yaml -f ../../model-engine-internal/model-engine-internal/resources/values/values_training.yaml 
```

### DIFF
--- a/charts/model-engine/templates/istio-metrics.yaml
+++ b/charts/model-engine/templates/istio-metrics.yaml
@@ -1,7 +1,7 @@
 apiVersion: telemetry.istio.io/v1alpha1
 kind: Telemetry
 metadata:
-  name: custom-tags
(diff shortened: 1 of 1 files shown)


yunfeng-scale (approved):

yunfeng-scale:
is there a way to ignore creating these objects for test builds?

song-william:
> is there a way to ignore creating these objects for test builds?

We can, but I think we actually do want to create these objects so route level istio-metrics are tracked for the test deployments.

PR #375, opened 2023-11-14
Forward HTTP status code for sync requests

yunfeng-scale:
# Pull Request Summary

We used to always forward with 200s ignoring downstream HTTP codes, now forwarding the code as well

## Test Plan and Usage Guide

unit tests and tested with echo server to make sure 500 status was forwarded

```
curl -X POST localhost:5000/predict -d '{"args":{}}' -H "content-type:application/json" -vv
Note: Unnecessary use of -X or --request, POST is already inferred.
* processing: localhost:5000/predict
*   Trying [::1]:5000...
* Connected to localhost (::1) port 5000
> POST /predict HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.68.0
> Accept: */*
> content-type:application/json
> Content-Length: 11
> 
< HTTP/1.1 500 Internal Server Error
< date: Tue, 14 Nov 2023 21:23:58 GMT
< server: uvicorn
< content-length: 15
< content-type: application/json
< 
* Connection #0 to host localhost left intact
{"result":"{}"}
```

### DIFF
--- a/model-engine/model_engine_server/inference/configs/service--forwarder-runnable-img-converted-from-artifact.yaml
+++ b/model-engine/model_engine_server/inference/configs/service--forwarder-runnable-img-converted-from-artifact.yaml
@@ -9,6 +9,7 @@ forwarder:
     model_engine_unwrap: false
     serialize_results_as_string: false
     wrap_response: false
+    forward_http_status: true
   async:
     user_port: 5005
     user_hostname: "localhost"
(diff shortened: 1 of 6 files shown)


yixu34 (approved):
Nice! Let's make an announcement before deploying this, because even though this is backwards compatible (schema-wise) and is more correct, it may surprise people if their alarms start going off.

seanshi-scale:
> Nice! Let's make an announcement before deploying this, because even though this is backwards compatible (schema-wise) and is more correct, it may surprise people if their alarms start going off.

iiuc we'd either need to update forwarder images on everyone's existing deployments or have people update their deployments for the changes to be picked up?

PR #378, opened 2023-11-15
Celery autoscaler

squeakymouse:
# Pull Request Summary

Move over the Celery autoscaler

## Test Plan and Usage Guide

Installing a test deployment autoscaled an async endpoint from 0 to 1 pod

### DIFF
--- a/charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
+++ b/charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
@@ -2,7 +2,7 @@
 {{- $app := include "modelEngine.celeryautoscalername" . }}
 {{- $env := .Values.context }}
 {{- $tag := .Values.tag }}
-{{- $message_broker := .Values.celery_autoscaler.message_broker }}
+{{- $message_broker := .Values.celeryBrokerType }}
 {{- $num_shards := .Values.celery_autoscaler.num_shards }}
 {{- $broker_name := ternary "redis-elasticache-message-broker-master" "sqs-message-broker-master" (eq $message_broker "elasticache") }}
 apiVersion: apps/v1
--- a/charts/model-engine/values_sample.yaml
+++ b/charts/model-engine/values_sample.yaml
@@ -92,11 +92,9 @@ autoscaling:
     enabled: false
 
 # for async endpoints, Celery autoscaler scales the number of pods based on number of requests
-# message_broker should be either "sqs" or "elasticache"
 # num_shards is number of instances of the autoscaler
 celery_autoscaler:
   enabled: true
-  message_broker: sqs
   num_shards: 3
 
 podDisruptionBudget:
@@ -247,5 +245,5 @@ imageCache:
           operator: "Exists"
           effect: "NoSchedule"
 
-# celeryBrokerType specifies the celery broker type for async endpoints (coming soon)
+# celeryBrokerType specifies the celery broker type for async endpoints, either "sqs" or "elasticache"
 celeryBrokerType: sqs
--- a/charts/model-engine/templates/_helpers.tpl
+++ b/charts/model-engine/templates/_helpers.tpl
@@ -32,9 +32,9 @@ If release name contains chart name it will be used as a full name.
 
 {{- define "modelEngine.celeryautoscalername" -}}
 {{- if .Values.serviceIdentifier }}
-{{- printf "celery-autoscaler-%s-%s" .Values.celery_autoscaler.message_broker .Values.serviceIdentifier }}
+{{- printf "celery-autoscaler-%s-%s" .Values.celeryBrokerType .Values.serviceIdentifier }}
 {{- else }}
-{{- printf "celery-autoscaler-%s" .Values.celery_autoscaler.message_broker }}
+{{- printf "celery-autoscaler-%s" .Values.celeryBrokerType }}
 {{- end }}
 {{- end }}
 


charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
```diff
@@ -0,0 +1,96 @@
...
+        node-lifecycle: normal
+      tolerations:
+        - key: CriticalAddonsOnly
+          operator: Equal
+          value: 'true'
+          effect: NoSchedule
+      serviceAccountName: ml-k8s-admin
+      volumes:
+      - configMap:
+          name: ml-worker-config
```
yunfeng-scale, line 93: `{{ .Values.aws.configMap.name }}` ?

charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
```diff
@@ -0,0 +1,96 @@
...
+        - name: NUM_SHARDS
+          value: '{{ $num_shards }}'
+        image: "{{ $.Values.image.gatewayRepository }}:{{ $tag }}"
+        imagePullPolicy: Always
+        name: main
+        resources:
+          requests:
+            cpu: 1000m
+        volumeMounts:
+        - mountPath: /root/.aws/config
```
saiatmakuri, line 80: /opt?

model-engine/model_engine_server/core/celery_autoscaler.py
```diff
@@ -0,0 +1,635 @@
...
+
+
+ELASTICACHE_REDIS_BROKER = "redis-elasticache-message-broker-master"
+SQS_BROKER = "sqs-message-broker-master"
+
+UPDATE_DEPLOYMENT_MAX_RETRIES = 10
+
+SQS_SAMPLE_COUNT = 10
+
+logger = make_logger("model_engine_server.core.celery_autoscaler")
```
saiatmakuri, line 51: nit: use `logger_name()` that you can import as well

model-engine/model_engine_server/core/celery_autoscaler.py
```diff
@@ -0,0 +1,635 @@
...
+
+
+def excluded_namespaces():
+    try:
+        from plugins.celery_autoscaler_dependencies import CELERY_AUTOSCALER_EXCLUDED_NAMESPACES
+
+        return CELERY_AUTOSCALER_EXCLUDED_NAMESPACES
+    except ModuleNotFoundError:
+        return []
+    finally:
```
saiatmakuri, line 40: nit: unnecessary

model-engine/model_engine_server/core/celery_autoscaler.py
```diff
@@ -0,0 +1,635 @@
...
+            for db_index in get_all_db_indexes()
+        }
+    elif broker_type == "sqs":
+        # for sqs we will get active/reserved counts directly from sqs as opposed to using
+        #   an inspect object
+        inspect = {}
+    else:
+        raise ValueError("broker_type not redis or sqs, how did we get here?")
+
+    env = os.getenv("DD_ENV", "training")
```
saiatmakuri, line 561: do we want to default to training here?

model-engine/model_engine_server/core/celery_autoscaler.py
```diff
@@ -0,0 +1,635 @@
+import asyncio as aio
```
saiatmakuri, line 1: could this be moved to the `celery/` folder?

model-engine/model_engine_server/core/celery/celery_autoscaler.py
```diff
@@ -0,0 +1,633 @@
...
+
+            if deployment_broker != autoscaler_broker:
+                logger.debug(
+                    f"Skipping deployment {deployment_name}; deployment's broker {deployment_broker} is not {autoscaler_broker}"
+                )
+                continue
+
+            for f in dataclasses.fields(CeleryAutoscalerParams):
+                k = f.name
+                # TODO: this technically can remain the same, since Celery deployments won't have the `channel` field and
```
seanshi-scale, line 104: I guess we can remove this comment probably? there is no NATS autoscaler anymore

charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
```diff
@@ -0,0 +1,98 @@
+{{- if .Values.celery_autoscaler.enabled }}
+{{- $env := .Values.context }}
+{{- $tag := .Values.tag }}
+{{- $service_identifier := .Values.serviceIdentifier }}
+{{- $num_shards := ternary 30 3 (eq .Values.context "prod") }}
+{{- range $base_app := tuple "celery-autoscaler-elasticache" "celery-autoscaler-sqs" }}
```
seanshi-scale, line 6: it's probably fine to have this range but I think that we'd be able to choose which one based on whether we're having endpoints use redis or sqs at some point

charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
```diff
@@ -0,0 +1,98 @@
...
+{{- $app := ternary (printf "%s-%s" $base_app $service_identifier) $base_app (not (empty $service_identifier)) }}
+{{- $broker_name := ternary "redis-elasticache-message-broker-master" "sqs-message-broker-master" (eq $base_app "celery-autoscaler-elasticache") }}
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    app: {{ $app }}
+    team: infra
```
yixu34, line 15: These tags shouldn't be hardcoded, they're internal.

yixu34 (commented):
Let's parameterize the cost tags.

charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
yunfeng-scale: see other deployments. something like
```
    {{- include "modelEngine.selectorLabels.celeryAutoScaler" . | nindent 4 }}
    {{- include "modelEngine.labels" . | nindent 4 }}
    tags.datadoghq.com/service: {{ include "modelEngine.celeryAutoScalerName" . }}
```
should

charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
```diff
@@ -0,0 +1,98 @@
...
+  selector:
+    matchLabels:
+      app: {{ $app }}
+  template:
+    metadata:
+      annotations:
+        ad.datadoghq.com/main.logs: '[{"service": "{{ $app }}", "source": "python"}]'
+        sidecar.istio.io/inject: "false"
+      labels:
+        app: {{ $app }}
```
yunfeng-scale, line 33: these tags should also be templatized

charts/model-engine/templates/celery_autoscaler_stateful_set.yaml
yunfeng-scale: yeah i think that file should parametrized tags. don't see why balloon deployment needs a different `product` tag

charts/model-engine/values_sample.yaml
```diff
@@ -91,6 +91,14 @@ autoscaling:
   prewarming:
     enabled: false
 
+# for async endpoints, Celery autoscaler scales the number of pods based on number of requests
+# message_broker should be either "sqs" or "elasticache"
+# num_shards is number of instances of the autoscaler
+celery_autoscaler:
+  enabled: true
+  message_broker: sqs
```
seanshi-scale, line 99: this is already defined as the root-level "messageBrokerType" (can see that at the bottom of this file)

seanshi-scale (commented):
had one nit about deduplicating info in the helm values but the rest looks good

yunfeng-scale (approved):

PR #383, opened 2023-11-16
Fix integration test

yunfeng-scale:
# Pull Request Summary

Pull policy for these images should be IfNotPresent, otherwise minikube wouldn't be able to load them and fine tune jobs can't start. i'm not sure how finetune integration tests have been passing (maybe we actually pulled the image from minikube?)

## Test Plan and Usage Guide

### DIFF
--- a/charts/model-engine/templates/service_template_config_map.yaml
+++ b/charts/model-engine/templates/service_template_config_map.yaml
@@ -572,7 +572,7 @@ data:
                 {{- tuple $env_var | toYaml | nindent 16 }}
                 {{- end }}
                 {{- end }}
-              imagePullPolicy: Always
+              imagePullPolicy: IfNotPresent
               command:
                 - dumb-init
                 - --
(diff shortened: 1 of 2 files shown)


charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -572,7 +572,7 @@ data:
                 {{- tuple $env_var | toYaml | nindent 16 }}
                 {{- end }}
                 {{- end }}
-              imagePullPolicy: Always
+              imagePullPolicy: IfNotPresent
```
yunfeng-scale, line 575: @seanshi-scale let me know this change is okay

seanshi-scale: This should be fine, think our image tags are immutable

seanshi-scale (approved):

seanshi-scale:
iirc for the endpoints we explicitly load images into minikube at some point, but I'm not sure if the imagePullPolicy on the endpoints is the same

PR #385, opened 2023-11-17
Some updates to integration tests

yunfeng-scale:
# Pull Request Summary

Some updates to integration tests

## Test Plan and Usage Guide

### DIFF
--- a/integration_tests/test_endpoints.py
+++ b/integration_tests/test_endpoints.py
@@ -2,7 +2,6 @@ import asyncio
 import time
 
 import pytest
-from model_engine_server.common.env_vars import CIRCLECI
 from tenacity import RetryError, retry, retry_if_exception_type, stop_after_attempt, wait_fixed
 
 from .rest_api_utils import (
--- a/integration_tests/test_endpoints.py
+++ b/integration_tests/test_endpoints.py
@@ -237,7 +237,9 @@ def test_sync_streaming_model_endpoint(capsys):
             delete_model_endpoint(create_endpoint_request["name"], user)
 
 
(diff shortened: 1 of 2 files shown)


integration_tests/test_endpoints.py
```diff
@@ -237,7 +236,9 @@ def test_sync_streaming_model_endpoint(capsys):
             delete_model_endpoint(create_endpoint_request["name"], user)
 
 
-@pytest.mark.skipif(CIRCLECI, reason="skip on circleci since need to figure out s3 access")
+@pytest.mark.skipif(
```
yunfeng-scale, line 239: @saiatmakuri FYI, i don't think this sounds as an integration test. can consider checking # tokens from responses?

saiatmakuri: yeah can skip for now. essentially want to check if every tokenizer can be made before each deploy

integration_tests/test_fine_tunes.py
```diff
@@ -19,6 +20,10 @@
 MAX_RETRIES = 10
 
 
+@pytest.mark.skipif(
+    os.getenv("ENV") == "prod",
```
saiatmakuri, line 24: skip if == "launch" as well?maybe we can limit this to circleci for now

saiatmakuri (approved):

integration_tests/test_endpoints.py
yunfeng-scale: it currently fails for me since initdb is not in bash container. can be fixed by adding postgres there

integration_tests/test_fine_tunes.py
yunfeng-scale: training cluster is proper set up for this. i do think we want some e2e test like this.

PR #386, opened 2023-11-17
Revert "Found a bug in the codellama vllm model_len logic."

yunfeng-scale:
Reverts scaleapi/llm-engine#380


ian-scale (approved):

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -181,9 +181,9 @@
     "mammoth-coder": {"max_model_len": 16384, "max_num_batched_tokens": 16384},
     # Based on config here: https://huggingface.co/TIGER-Lab/MAmmoTH-Coder-7B/blob/main/config.json#L12
     # Can also see 13B, 34B there too
-    "codellama": {"max_model_len": 16384, "max_num_batched_tokens": 16384},
+    "code-llama": {"max_model_len": 16384, "max_num_batched_tokens": 16384},
```
ian-scale, line 184: quick question on this Yunfeng: I wanted the models to be named `codellama` to be in keeping with the way that it's formatted for finetuning, why are we switching?

PR #388, opened 2023-11-21
Don't install Celery autoscaler for test deployments

squeakymouse:
# Pull Request Summary

Since they might clash with the actual autoscaler

## Test Plan and Usage Guide


seanshi-scale (approved):
guess if we wanted we could add something to the values to force install celery autoscaler and propagate that to the justfile, but can do that later

PR #390, opened 2023-11-27
update tensor-rt llm in enum

ian-scale:
# Pull Request Summary

_What is this PR changing? Why is this change being made? Any caveats you'd like to highlight? Link any relevant documents, links, or screenshots here if applicable._
adds tensorrt-llm to enum for valid inference frameworks. currently `Model.list()` is broken due to the addition of endpoints that use tensorrt-llm as an inference framework when this framework isn't an option in the client. 

Ex: 

```
ValidationError: 1 validation error for ListLLMEndpointsResponse
model_endpoints -> 13 -> inference_framework
  value is not a valid enumeration member; permitted: 'deepspeed', 'text_generation_inference', 'vllm', 'lightllm' (type=type_error.enum; enum_values=[<LLMInferenceFramework.DEEPSPEED: 'deepspeed'>, <LLMInferenceFramework.TEXT_GENERATION_INFERENCE: 'text_generation_inference'>, <LLMInferenceFramework.VLLM: 'vllm'>, <LLMInferenceFramework.LIGHTLLM: 'lightllm'>])
```

## Test Plan and Usage Guide

_How did you validate that your PR works correctly? How do you run or demo the code? Provide enough detail so a reviewer can reasonably reproduce the testing procedure. Paste example command line invocations if applicable._

PR #396, opened 2023-11-29
fixing tensorrt-llm enum value (fixes #390)

ian-scale:
# Pull Request Summary

This pr fixes a mistake introduced in https://github.com/scaleapi/llm-engine/pull/390 where I mistakenly used a `-` instead of a `_` for the inference framework. merging the pr and bumping the pypi client will fix pydantic `Model.list()` issues.

## Test Plan and Usage Guide

Local deployment of the client installed with `pip install -e .` and then running `Model.list()` works without issue.
Screenshot 2023-11-29 at 10 57 36 AM

### DIFF
--- a/clients/python/llmengine/__init__.py
+++ b/clients/python/llmengine/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.0b20"
+__version__ = "0.0.0b21"
 
 import os
 from typing import Sequence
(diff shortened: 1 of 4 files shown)


yixu34:
Ok now that we've been burned by this, let's please add a test in a follow-up PR: https://github.com/scaleapi/llm-engine/pull/390#pullrequestreview-1751155426

Maybe we can have a parameterized unit test that loops over the enum values and has fake objects in there? I guess this is partly due to Python's dynamic nature.

PR #397, opened 2023-11-30
time completions use case

saiatmakuri:
# Pull Request Summary

the use case for completions (sync and stream) includes preprocessing + forwarding request to the model + postprocessing. time the entire flow for total tokens per second

## Test Plan and Usage Guide

test deployment


seanshi-scale (approved):
nice more data

PR #399, opened 2023-12-05
Add MultiprocessingConcurrencyLimiter to gateway

squeakymouse:
# Pull Request Summary

Add concurrency limiter so gateway returns 429s instead of 503s when overloaded

Remove kubernetes liveness probe to prevent pods from restarting when under high load and the healthcheck fails (purpose of the liveness probe is to restart in case of deadlocks? which doesn't seem relevant for us)

Increase Uvicorn worker concurrency now that we have rate limiting (the latter should take precedence so that we don't return 503s)

## Test Plan and Usage Guide

Load tested with https://github.com/rakyll/hey on a test deployment and saw 429s instead of 503s for concurrency up to 5000

### DIFF
--- a/model-engine/model_engine_server/api/app.py
+++ b/model-engine/model_engine_server/api/app.py
@@ -40,12 +40,15 @@ concurrency_limiter = MultiprocessingConcurrencyLimiter(
     concurrency=MAX_CONCURRENCY, fail_on_concurrency_limit=True
 )
 
+healthcheck_routes = ["/healthcheck", "/healthz", "/readyz"]
+
 
 class CustomMiddleware(BaseHTTPMiddleware):
     async def dispatch(self, request: Request, call_next):
         try:
             LoggerTagManager.set(LoggerTagKey.REQUEST_ID, str(uuid.uuid4()))
-            if request.url.path in ["/healthcheck", "/healthz", "/readyz"]:
+            # we intentionally exclude healthcheck routes from the concurrency limiter
+            if request.url.path in healthcheck_routes:
                 return await call_next(request)
             with concurrency_limiter:
                 return await call_next(request)
@@ -109,10 +112,10 @@ def load_redis():
     get_or_create_aioredis_pool()
 
 
-# these routes should match those exempt from the concurrency limiter in the middleware
-@app.get("/healthcheck")
-@app.get("/healthz")
-@app.get("/readyz")
 def healthcheck() -> Response:
     """Returns 200 if the app is healthy."""
     return Response(status_code=200)
+
+
+for endpoint in healthcheck_routes:
+    app.get(endpoint)(healthcheck)


model-engine/model_engine_server/common/concurrency_limiter.py
```diff
@@ -0,0 +1,35 @@
...
+                not fail_on_concurrency_limit
+            )  # we want to block if we want to queue up requests
+        else:
+            self.semaphore = None
+            self.blocking = False  # Unused
+
+    def __enter__(self):
+        logger.debug("Entering concurrency limiter semaphore")
+        if self.semaphore and not self.semaphore.acquire(block=self.blocking):
+            logger.warning("Too many requests, returning 429")
```
ian-scale, line 27: nit: can we log the # of requests instead of this log? i.e. # of requests that exceed the concurrency limit

seanshi-scale: any requests over the concurrency limit will immediately be returned as 429, so we'd always be logging whatever the value of `MAX_CONCURRENCY` is (which still seems good to log though)

seanshi-scale (commented):
code seems good, but I would like more complete load test results (e.g. some stripped-down version of the load test docs phil and I have written) since the change we're making is quite dependent on stuff outside of this code, and there's nontrivial interactions between this code, whatever istio setup we have, where the requests are coming from, etc. 

E.g. one reason you might be seeing different behavior with `just up` and when deployed on a test deployment are because there may be request queues hidden somewhere for your test deployment (e.g. there's a bunch of other stuff in between your devbox and the gateway pods, like some istio layer for sure, the istio-proxy pods), as opposed to when testing locally

ian-scale:
@squeakymouse is this rate-limiting per-user or is it system-wide? I think something we could consider working towards in the future would be implementing per-user rate limits - I could foresee this being awkward if someone sends their first request ever at a time when system load is high and gets back a 429. 

Actually, the best case might be instituting per-user rate limits and system rate limits. This way, no one user can monopolize all our throughput at any point in time. What do you think? 

Also, I could be totally missing a line in the code and the per-user rate limits have already been added.

model-engine/model_engine_server/api/app.py
```diff
@@ -32,12 +33,29 @@
...
+    concurrency=MAX_CONCURRENCY, fail_on_concurrency_limit=True
+)
+
 
 class CustomMiddleware(BaseHTTPMiddleware):
     async def dispatch(self, request: Request, call_next):
         try:
-            LoggerTagManager.set(LoggerTagKey.REQUEST_ID, str(uuid.uuid4()))
-            return await call_next(request)
+            with concurrency_limiter:
```
song-william, line 47: What do people think about trying this out with just a specific route at first? Looking at the breakdown in the [past week](https://app.datadoghq.com/metric/explorer?start=1701146189478&end=1701750989478&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VK7ueJiWyDYgGGgQGGEgACo4CFEABACiAG5wIhYGIEGqIRjKpjpYZCA8ALpUpf5hEVCeGE4a+NoAxkrIIOoIuh2WVCIY6iVm2uV5dTQ9GniqyBAaEjgAdL6Ma0alK2sd+DlrCElRwB1oeCs6TiPqyKpoALR+GaIwTyJwYCkUKQIrUSeSQiJwYeTyB44ATPbIwURwP4AQh2ZhWThwim0wKkyDwIgA1iIcGARDwUgAjLApYCo8wYrE4slrWJOA74hAACgAlJUaiAVlg0N1QPIoogEN0QFAcDA2aENBAOnlol8nHJFMo2tEoCqEU56ExlMT3FFKvxlvJMFh1QopRSMNpLNUeHwBVbxABhKTCGAoEQhNA8IA), `get_/v1/async-tasks/_task_id` is the most common route by far. 

I'm not sure if it make sense to do a global limit, as we know some routes take more time than others.

model-engine/model_engine_server/common/concurrency_limiter.py
```diff
@@ -0,0 +1,35 @@
+from multiprocessing import BoundedSemaphore
```
song-william, line 1: I'm wonder if we should be using an `async` semaphore rather than a multiprocessing one, since FastAPI is using async. 

This [SO comment](https://stackoverflow.com/questions/69698479/python-threading-semaphore-vs-asyncio-semaphore) seems to suggest that we should be using the corresponding semaphore:
> use the correct type of semaphore for the form of concurrency being used

saiatmakuri:
curious if there was consideration of 3P libraries, like [slowapi](https://github.com/laurentS/slowapi), where rate limiting can be extended to redis, time-bounded, by user strategy, easy customization of limits per route, etc

model-engine/model_engine_server/api/app.py
```diff
@@ -91,6 +109,7 @@ def load_redis():
     get_or_create_aioredis_pool()
 
 
+# these routes should match those exempt from the concurrency limiter in the middleware
```
song-william, line 112: To better link the routes between here <> middleware with code, we could define a shared list variable. Something like this could work. 

```
health_routes = ["/healthcheck", "/healthz", "/readyz"]

def healthcheck() -> Response:
    """Returns 200 if the app is healthy."""
    return Response(status_code=200)

for endpoint in health_routes:
    app.get(endpoint)(healthcheck)
```

model-engine/model_engine_server/api/app.py
```diff
@@ -32,12 +33,31 @@
...
+    concurrency=MAX_CONCURRENCY, fail_on_concurrency_limit=True
+)
+
 
 class CustomMiddleware(BaseHTTPMiddleware):
     async def dispatch(self, request: Request, call_next):
         try:
             LoggerTagManager.set(LoggerTagKey.REQUEST_ID, str(uuid.uuid4()))
-            return await call_next(request)
+            if request.url.path in ["/healthcheck", "/healthz", "/readyz"]:
```
song-william, line 48: nit: would just add a comment here that we intentionally exclude health check routes from the concurrency limiter.

song-william (approved):
Thanks!

PR #402, opened 2023-12-07
fix completions request id

saiatmakuri:
# Pull Request Summary

have completions use the same request id as that was generated by the middleware

## Test Plan and Usage Guide

local

### DIFF
--- a/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
@@ -9,7 +9,6 @@ import math
 import os
 from dataclasses import asdict
 from typing import Any, AsyncIterable, Dict, List, Optional, Union
-from uuid import uuid4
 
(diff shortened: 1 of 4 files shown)


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1448,7 +1452,8 @@ async def execute(
             ObjectNotAuthorizedException: If the owner does not own the model endpoint.
         """
 
-        request_id = str(uuid4())
+        request_id = LoggerTagManager.get(LoggerTagKey.REQUEST_ID)
+        assert request_id is not None
```
yixu34, line 1456: Actually, do we want things to blow up if we can't get a request ID? That might be a bit harsh.

saiatmakuri: unit tests agree

yixu34 (approved):

PR #403, opened 2023-12-08
Allow latest inference framework tag

squeakymouse:
# Pull Request Summary

Allow option "latest" for inference framework tag to grab the most recently pushed image in ECR

## Test Plan and Usage Guide

Worked locally with `just up`, currently making a test deployment to test with

### DIFF
--- a/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
@@ -9,7 +9,6 @@ import math
 import os
 from dataclasses import asdict
 from typing import Any, AsyncIterable, Dict, List, Optional, Union
-from uuid import uuid4
 
 from model_engine_server.common.config import hmi_config
 from model_engine_server.common.dtos.llms import (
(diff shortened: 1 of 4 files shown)


yunfeng-scale (approved):

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1102,12 +1112,18 @@ async def execute(
             llm_metadata = (model_endpoint.record.metadata or {}).get("_llm", {})
             inference_framework = llm_metadata["inference_framework"]
 
+            if request.inference_framework_image_tag == "latest":
```
saiatmakuri, line 1115: since we convert "latest" to a specific image tag in the create route, we just want to fetch that tag right? if you fetch "latest" again here, we might so the  a later version that the model was not created with?

saiatmakuri: ah I meant this scenario
- Dec 1: latest --> v1, user creates model
- Dec 2: latest --> v2
- Dec 3: user gets model, tag on their model would be v1 not v2

saiatmakuri (approved):

PR #414, opened 2023-12-26
handle update endpoint errors

saiatmakuri:
# Pull Request Summary

- throw a 400 for `ObjectHasInvalidValueException` and `EndpointResourceInvalidRequestException` for update model endpoint route
- capture `EndpointInfraStateNotFound` as an explicit 500

## Test Plan and Usage Guide

update unit tests

### DIFF
--- a/model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
@@ -364,11 +364,21 @@ class UpdateModelEndpointByIdV1UseCase:
         # from infra_state to avoid passing in None to validate_resource_requests.
         validate_resource_requests(
             bundle=bundle,
-            cpus=request.cpus or infra_state.resource_state.cpus,
-            memory=request.memory or infra_state.resource_state.memory,
-            storage=request.storage or infra_state.resource_state.storage,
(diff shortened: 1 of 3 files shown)


model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
```diff
@@ -362,13 +362,22 @@ async def execute(
         # infra_state to make sure that after the update, all resources are valid and in sync.
         # E.g. If user only want to update gpus and leave gpu_type as None, we use the existing gpu_type
         # from infra_state to avoid passing in None to validate_resource_requests.
+        raw_request = request.dict(exclude_unset=True)
```
song-william, line 365: Good Catch here. I'm guessing the bug here was that pydantic fields with non-falsey defaults were evaluating as being set by the user even if they were not explicitly set in the request?

song-william (approved):

model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
saiatmakuri: yeah, primarily that a user could not specify `None` for fields for which it's allowed, ie `gpu_type`

PR #415, opened 2023-12-31
enable sensitive log mode

song-william:
# Pull Request Summary
Enable sensitive log mode for use cases that want avoid logging user-specific data (e.g prompts and responses)

## Test Plan and Usage Guide
Deploy in training and confirmed in datadog that logs no longer show up in vllm or logged by our gateway.

### DIFF
--- a/model-engine/tests/unit/api/test_llms.py
+++ b/model-engine/tests/unit/api/test_llms.py
@@ -156,6 +156,9 @@ def test_completion_sync_endpoint_not_found_returns_404(
     assert response_1.status_code == 404
 
 
+# When enabling this test, other tests fail with "RunTumeError got Future <Future pending> attached to a different loop"
+# https://github.com/encode/starlette/issues/1315#issuecomment-980784457
+@pytest.mark.skip(reason="Need to figure out FastAPI test client asyncio funkiness")
 def test_completion_stream_success(
     llm_model_endpoint_streaming: ModelEndpoint,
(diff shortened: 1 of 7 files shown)


saiatmakuri (approved):

model-engine/tests/unit/api/test_llms.py
```diff
@@ -156,6 +156,8 @@ def test_completion_sync_endpoint_not_found_returns_404(
     assert response_1.status_code == 404
 
 
+# When enabling this test, other tests fail with "RunTumeError got Future <Future pending> attached to a different loop"
+# https://github.com/encode/starlette/issues/1315#issuecomment-980784457
```
song-william, line 160: To reproduce, run
```
GIT_TAG=$(git rev-parse HEAD) WORKSPACE=..  pytest -v -s tests/unit/api/test_tasks.py::test_create_streaming_task_success tests/unit/api/test_llms.py::test_completion_stream_success
```

When running these tests individually, we don't have an issue. Only when running both does this error appear. I put `id(asyncio.get_running_loop())` throughout the `create_completion_stream_task` method and everything lined up and didn't error, so there must be something up with the `TestClient` library itself. Time-boxing this for now.

song-william: I tried the small fixes in https://github.com/encode/starlette/issues/1315#issuecomment-980784457, none of them seemed to work. I think we will either need to downgrade `fastapi` as folks suggested.

PR #425, opened 2024-01-23
Azure refactor for async endpoints

squeakymouse:
# Pull Request Summary

Azure support for a very limited subset of functionality; the only one needed right now is async endpoint creation and inference

## Test Plan and Usage Guide

Deployed to an Azure test cluster

### DIFF
--- a/model-engine/model_engine_server/api/dependencies.py
+++ b/model-engine/model_engine_server/api/dependencies.py
@@ -44,6 +44,12 @@ from model_engine_server.domain.services import (
     LLMModelEndpointService,
     ModelEndpointService,
 )
+from model_engine_server.inference.domain.gateways.streaming_storage_gateway import (
+    StreamingStorageGateway,
+)
+from model_engine_server.inference.infra.gateways.firehose_streaming_storage_gateway import (
+    FirehoseStreamingStorageGateway,
+)
 from model_engine_server.infra.gateways import (
     ABSFileStorageGateway,
     ABSFilesystemGateway,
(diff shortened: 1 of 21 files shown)


model-engine/model_engine_server/db/base.py
```diff
@@ -42,17 +46,28 @@ def get_engine_url(env: Optional[str] = None, read_only: bool = True, sync: bool
...
-        engine_url = f"postgresql://{user}:{password}@{host}:{port}/{dbname}"
+        if infra_config().cloud_provider == "azure":
+            client = SecretClient(
+                vault_url=f"https://{os.environ.get('KEYVAULT_NAME')}.vault.azure.net",
+                credential=ManagedIdentityCredential(
+                    client_id=os.getenv("AZURE_KEYVAULT_IDENTITY_CLIENT_ID")
+                ),
+            )
+            engine_url = client.get_secret(key_file).value
+            logger.info(f"Connecting to db {engine_url}")
```
yunfeng-scale, line 58: i think this would expose password?

.circleci/config.yml
```diff
@@ -108,13 +108,13 @@ jobs:
             cd $HOME
             curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube_latest_amd64.deb
             sudo dpkg -i minikube_latest_amd64.deb
-            minikube start --vm-driver=docker --kubernetes-version=v1.23.0 --memory=49152 --cpus=14
+            minikube start --vm-driver=docker --kubernetes-version=v1.27.7 --memory=49152 --cpus=14
```
yunfeng-scale, line 111: hmm, would this require a cluster update for production clusters?

charts/model-engine/values_sample.yaml
```diff
@@ -139,6 +142,8 @@ serviceTemplate:
 config:
   values:
     infra:
+      # cloud_provider [required]; either "aws" or "azure"
+      cloud_provider: aws
       # k8s_cluster_name [required] is the name of the k8s cluster
```
yunfeng-scale, line 147: some of the config values seem to be only used with aws, like `default_region`, `ml_account_id`, `s3_bucket`, sqs stuff. not need to be done in this PR, but we should move them into aws config section

yunfeng-scale (approved):
great work!

PR #427, opened 2024-01-26
LLM benchmark script improvements

seanshi-scale:
# Pull Request Summary

Add some changes to the benchmark script to be more detailed

## Test Plan and Usage Guide

Ran benchmark script locally


scripts/throughput_benchmarks.py
```diff
@@ -280,12 +286,17 @@ def run_benchmark(
         "avg_per_session_sampling_throughput": num_sampled_tokens
         / (elapsed * avg_completion_time / (avg_prefill_time + avg_completion_time))
         / concurrency,
+        "avg_request_throughput": n / elapsed,
         "avg_inter_token_latency": sum(inter_token_latency) / n,
```
yunfeng-scale, line 290: what about percentiles for this? also i wonder if we can have percentiled numbers for throughput numbers?

yunfeng-scale (approved):

scripts/throughput_benchmarks.py
yixu34: +1 I think throughput is more easily interpretable to users?

seanshi-scale: I've added percentiles for inter-token latency + time to first token, not sure how to get percentiles for throughput (since I think it's a single value for the entire test)

PR #428, opened 2024-02-01
Logging post inference hook implementation

tiffzhao5:
# Pull Request Summary

Logging post inference hook to firehose stream. 

## Test Plan and Usage Guide

Runs successfully locally. Load test of 1000 requests success (finished in < 1min and all show up in table).

### DIFF
--- a/model-engine/model_engine_server/common/constants.py
+++ b/model-engine/model_engine_server/common/constants.py
@@ -3,7 +3,6 @@ from pathlib import Path
 BILLING_POST_INFERENCE_HOOK: str = "billing"
 CALLBACK_POST_INFERENCE_HOOK: str = "callback"
 LOGGING_POST_INFERENCE_HOOK: str = "logging"
-LOGGING_FIREHOSE_STREAM: str = "PUT-ML-LOGGING-TEST-1"
 READYZ_FPATH: str = "/tmp/readyz"
 DEFAULT_CELERY_TASK_NAME: str = "hosted_model_inference.inference.async_inference.tasks.predict"
 LIRA_CELERY_TASK_NAME: str = "ml_serve.celery_service.exec_func"
--- a/model-engine/model_engine_server/common/constants.py
+++ b/model-engine/model_engine_server/common/constants.py
@@ -2,6 +2,8 @@ from pathlib import Path
 
 BILLING_POST_INFERENCE_HOOK: str = "billing"
 CALLBACK_POST_INFERENCE_HOOK: str = "callback"
+LOGGING_POST_INFERENCE_HOOK: str = "logging"
+LOGGING_FIREHOSE_STREAM: str = "PUT-ML-LOGGING-TEST-1"
 READYZ_FPATH: str = "/tmp/readyz"
 DEFAULT_CELERY_TASK_NAME: str = "hosted_model_inference.inference.async_inference.tasks.predict"
 LIRA_CELERY_TASK_NAME: str = "ml_serve.celery_service.exec_func"
--- a/model-engine/model_engine_server/inference/async_inference/tasks.py
+++ b/model-engine/model_engine_server/inference/async_inference/tasks.py
@@ -46,6 +46,10 @@ def init_worker_global():
         default_callback_url=endpoint_config.default_callback_url,
         default_callback_auth=endpoint_config.default_callback_auth,
         monitoring_metrics_gateway=DatadogInferenceMonitoringMetricsGateway(),
+        endpoint_id=endpoint_config.endpoint_id,
+        endpoint_type=endpoint_config.endpoint_type,
+        bundle_id=endpoint_config.bundle_id,
+        labels=endpoint_config.labels,
     )
     # k8s health check
     with open(READYZ_FPATH, "w") as f:
--- a/model-engine/model_engine_server/inference/async_inference/tasks.py
+++ b/model-engine/model_engine_server/inference/async_inference/tasks.py
@@ -18,6 +18,9 @@ from model_engine_server.inference.common import (
 from model_engine_server.inference.infra.gateways.datadog_inference_monitoring_metrics_gateway import (
     DatadogInferenceMonitoringMetricsGateway,
 )
+from model_engine_server.inference.infra.gateways.firehose_streaming_storage_gateway import (
+    FirehoseStreamingStorageGateway,
+)
 from model_engine_server.inference.post_inference_hooks import PostInferenceHooksHandler
 
 logger = make_logger(logger_name())
@@ -50,6 +53,7 @@ def init_worker_global():
         endpoint_type=endpoint_config.endpoint_type,
         bundle_id=endpoint_config.bundle_id,
         labels=endpoint_config.labels,
+        streaming_storage_gateway=FirehoseStreamingStorageGateway(),
     )
     # k8s health check
     with open(READYZ_FPATH, "w") as f:
--- a/model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
+++ b/model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+from typing import Any, Dict
 
 
 class StreamingStorageGateway(ABC):
@@ -7,7 +8,7 @@ class StreamingStorageGateway(ABC):
     """
 
     @abstractmethod
-    def put_record(self, stream_name: str, record: str) -> None:
+    def put_record(self, stream_name: str, record: Dict[str, Any]) -> None:
         """
         Put a record into a streaming storage mechanism.
 
--- a/model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
+++ b/model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
@@ -1,0 +1,18 @@
+from abc import ABC, abstractmethod
+
+
+class StreamingStorageGateway(ABC):
+    """
+    Base class for a gateway that stores data through a streaming mechanism.
+    """
+
+    @abstractmethod
+    def put_record(self, stream_name: str, record: str) -> None:
+        """
+        Put a record into a streaming storage mechanism.
+
+        Args:
+            stream_name: The name of the stream.
+            record: The record to put into the stream.
+        """
+        pass
--- a/model-engine/model_engine_server/inference/infra/gateways/firehose_streaming_storage_gateway.py
+++ b/model-engine/model_engine_server/inference/infra/gateways/firehose_streaming_storage_gateway.py
@@ -1,0 +1,52 @@
+import boto3
+from model_engine_server.core.config import infra_config
+from model_engine_server.core.loggers import logger_name, make_logger
+from model_engine_server.inference.domain.gateways.streaming_storage_gateway import (
+    StreamingStorageGateway,
+)
+
+logger = make_logger(logger_name())
+
+
+class FirehoseStreamingStorageGateway(StreamingStorageGateway):
+    """
+    A gateway that stores data through the AWS Kinesis Firehose streaming mechanism.
+    """
+
+    def __init__(self):
+        """
+        Creates a new firehose client.
+
+        Streams with Snowflake as a destination live in the ml account while
+        ml-worker lives in the prod account. Firehose doesn't support resource-based
+        policies, so we need to assume a role in the ml account to write to the stream.
+        """
+        sts_client = boto3.client("sts", region_name=infra_config().default_region)
+        assumed_role_object = sts_client.assume_role(
+            RoleArn=infra_config().firehose_stream_logging_role_arn,
+            RoleSessionName="AssumeMlLoggingRoleSession",
+        )
+        credentials = assumed_role_object["Credentials"]
+        session = boto3.Session(
+            aws_access_key_id=credentials["AccessKeyId"],
+            aws_secret_access_key=credentials["SecretAccessKey"],
+            aws_session_token=credentials["SessionToken"],
+        )
+        firehose_client = session.client("firehose", region_name=infra_config().default_region)
+        self._firehose_client = firehose_client
+
+    def put_record(self, stream_name: str, record: str) -> None:
+        """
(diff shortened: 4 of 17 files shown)


model-engine/model_engine_server/inference/post_inference_hooks.py
```diff
@@ -76,6 +83,69 @@ def handle(
...
+            "ENDPOINT_TYPE": self._endpoint_type.value,
+            "BUNDLE_ID": self._bundle_id,
+            "LABELS": self._labels,
+        }
+        data = json.dumps(data_record)
+        firehose_response = firehose_client.put_record(
+            DeliveryStreamName=LOGGING_FIREHOSE_STREAM, Record={"Data": data.encode("utf-8")}
+        )
+        assert firehose_response["ResponseMetadata"]["HTTPStatusCode"] == 200
+        logger.info(f"Logged to firehose stream {LOGGING_FIREHOSE_STREAM}: {data_record}")
```
song-william, line 146: I might have missed some earlier PRs, but I think we want to follow Clean Arch here and put this "stream publishing" logic behind an interface. 

The montioring_gateway in this file is already an example: https://github.com/scaleapi/llm-engine/blob/112513cd46a50f4351e93e6e6648ad51c58d9b02/model-engine/model_engine_server/inference/post_inference_hooks.py#L90

model-engine/model_engine_server/inference/post_inference_hooks.py
```diff
@@ -76,6 +83,69 @@ def handle(
...
+            RoleArn=infra_config().firehose_role_arn,
+            RoleSessionName="AssumeRoleSession1",
+        )
+        credentials = assumed_role_object["Credentials"]
+        session = boto3.Session(
+            aws_access_key_id=credentials["AccessKeyId"],
+            aws_secret_access_key=credentials["SecretAccessKey"],
+            aws_session_token=credentials["SessionToken"],
+        )
+        firehose_client = session.client("firehose", region_name=infra_config().default_region)
```
song-william, line 120: okay, I think I got my wires crossed a bit earlier. Rather than requiring the service to assume a new IAM role, we should expand the existing role's perms to enable publishing to this stream. 

Effectively, we want to add the permissions declared [here](https://github.com/scaleapi/Terracode-ML/pull/596/files#diff-f2db50a56315df230a5a73d6935fa9f8f4700f61fe4ac724c034c491136333e1R655-R660) to the `ml-worker` role already assumed by endpoints. 

We can then follow the existing pattern for instantiating AWS clients ([s3 example](https://github.com/scaleapi/llm-engine/blob/14635b0be69c5331b90f2a50f263b57e9558a714/model-engine/model_engine_server/infra/gateways/s3_filesystem_gateway.py#L15)) inside the implementations.

song-william: I'm sorry I caught this later @tiffzhao5. We might need to redo some of the TF prs. We also probably don't need to introduce a new role ARN in this PR: https://github.com/scaleapi/models/pull/8808

model-engine/model_engine_server/inference/post_inference_hooks.py
song-william: I think this is the latest team guide on Clean Arch: https://docs.google.com/document/d/1ja-koN5EpBs-CnbVR0lYGgezBxw-Va2a6OWQqPgdpEs/edit#heading=h.lwro9snaxj7h

song-william (commented):
Some higher-level comments about roles and Clean arch. Let me know if you want to sync offline.

model-engine/model_engine_server/inference/post_inference_hooks.py
tiffzhao5: By adding the permissions to the `ml-worker` role already assumed by the endpoints, are you referring to this? https://github.com/scaleapi/Terracode-Infra/pull/2357/files because I've tried this already and it didn't work. This was because the firehose stream I created lives in the ml-account, while ml-worker lives in the scale-prod account, so the solution was to create a new IAM role in the ml-account (and serving) and have ml-worker assume that role.

Did I understand your proposal correctly? Also, I'm open to other solutions -- doing it this way was advised by platform so I went with it given little context on this.

model-engine/model_engine_server/inference/post_inference_hooks.py
tiffzhao5: I see, I think it makes sense architecturally for this to be in a separate gateway and I can work on this! This is my first PR for this btw so you didn't miss anything -- happy to iterate on this.

model-engine/model_engine_server/inference/post_inference_hooks.py
song-william: I think your original proposal and explanation is correct. Looks like Kinesis Firehose doesn't support `resource-based` policies ([link](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_aws-services-that-work-with-iam.html)). If it did, we would then just attach a policy to the firehose to allow cross-account `ml-worker` to publish to the stream.

Looks like your approach is required. I would just leave a comment in the code explaining why an additional role is required. 

In summary, our current setup is "service deployed in ML account assumes role from Prod account to assume a role in ML account to publish to ML account resources" 😿. Not a problem to fix in this PR.

song-william: Another option is to instead move the streams into the `Prod` account, but that probably isn't worth it just for IAM reasons.

model-engine/model_engine_server/inference/post_inference_hooks.py
```diff
@@ -76,6 +83,69 @@ def handle(
...
+    def handle(
+        self,
+        request_payload: EndpointPredictV1Request,
+        response: Dict[str, Any],
+        task_id: Optional[str],
+    ):
+        sts_client = boto3.client("sts", region_name=infra_config().default_region)
+        assumed_role_object = sts_client.assume_role(
+            RoleArn=infra_config().firehose_role_arn,
+            RoleSessionName="AssumeRoleSession1",
```
yunfeng-scale, line 112: use a more descriptive name?

model-engine/model_engine_server/common/constants.py
```diff
@@ -2,6 +2,8 @@
 
 BILLING_POST_INFERENCE_HOOK: str = "billing"
 CALLBACK_POST_INFERENCE_HOOK: str = "callback"
+LOGGING_POST_INFERENCE_HOOK: str = "logging"
+LOGGING_FIREHOSE_STREAM: str = "PUT-ML-LOGGING-TEST-1"
```
seanshi-scale, line 6: guessing this is just for testing/not used in production?

model-engine/model_engine_server/inference/post_inference_hooks.py
yunfeng-scale: agreed, the name `firehose` shouldn't appear in llm-engine.

model-engine/model_engine_server/inference/post_inference_hooks.py
saiatmakuri, line 120: nit: move logic into a helper that returns a `firehose_client`

model-engine/model_engine_server/inference/post_inference_hooks.py
saiatmakuri, line 146: saw that firehose_response should return a `RecordId` that could be usefull for logging

model-engine/model_engine_server/inference/post_inference_hooks.py
```diff
@@ -76,6 +83,69 @@ def handle(
...
+            "REQUEST_BODY": request_payload.json(),
+            "RESPONSE_BODY": response,
+            "ENDPOINT_ID": self._endpoint_id,
+            "ENDPOINT_NAME": self._endpoint_name,
+            "ENDPOINT_TYPE": self._endpoint_type.value,
+            "BUNDLE_ID": self._bundle_id,
+            "LABELS": self._labels,
+        }
+        data = json.dumps(data_record)
+        firehose_response = firehose_client.put_record(
```
saiatmakuri, line 142: if this errors, we'd be missing data right?

model-engine/model_engine_server/common/constants.py
saiatmakuri, line 6: nit: this should be in infra config

tiffzhao5: for production as well, will move this to infra config

model-engine/model_engine_server/inference/post_inference_hooks.py
tiffzhao5: yeah will add this condition in

model-engine/model_engine_server/inference/infra/gateways/firehose_streaming_storage_gateway.py
```diff
@@ -0,0 +1,56 @@
...
+        Put a record into a Firehose stream.
+
+        Args:
+            stream_name: The name of the stream.
+            record: The record to put into the stream.
+        """
+        firehose_response = self._get_firehose_client().put_record(
+            DeliveryStreamName=stream_name, Record={"Data": record.encode("utf-8")}
+        )
+        assert firehose_response["ResponseMetadata"]["HTTPStatusCode"] == 200
```
saiatmakuri, line 53: maybe change this to a custom exception name, eg add something like a `StreamPutException` [here](https://github.com/scaleapi/llm-engine/blob/main/model-engine/model_engine_server/domain/exceptions.py)

model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABC, abstractmethod
+
+
+class StreamingStorageGateway(ABC):
+    """
+    Base class for a gateway that stores data through a streaming mechanism.
+    """
+
+    @abstractmethod
+    def put_record(self, stream_name: str, record: str) -> None:
```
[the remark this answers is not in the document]
tiffzhao5: hm maybe dictionary is better, will change it

model-engine/model_engine_server/inference/post_inference_hooks.py
```diff
@@ -115,6 +195,8 @@ def handle(
     ):
         if isinstance(response, JSONResponse):
             loaded_response = json.loads(response.body)
+            if response.status_code >= 400:
```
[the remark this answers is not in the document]
tiffzhao5: I guess this is a way to catch errors. seems like callbacks still run even if the response errored out so just wanted logging to not happen

model-engine/model_engine_server/inference/async_inference/tasks.py
```diff
@@ -46,6 +49,11 @@ def init_worker_global():
         default_callback_url=endpoint_config.default_callback_url,
         default_callback_auth=endpoint_config.default_callback_auth,
         monitoring_metrics_gateway=DatadogInferenceMonitoringMetricsGateway(),
+        endpoint_id=endpoint_config.endpoint_id,
+        endpoint_type=endpoint_config.endpoint_type,
+        bundle_id=endpoint_config.bundle_id,
+        labels=endpoint_config.labels,
+        streaming_storage_gateway=FirehoseStreamingStorageGateway(),
```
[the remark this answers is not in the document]
tiffzhao5: yeah that could work

model-engine/model_engine_server/inference/post_inference_hooks.py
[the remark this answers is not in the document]
saiatmakuri: if we throw StreamPutException, we can explicitly catch that so that we're more explicit about errors thrown rather than any > 400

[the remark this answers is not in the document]
tiffzhao5: ah okay I misread your earlier comment in the previous PR review -- I thought you were referring to not logging data if the response itself errored out and not firehose's `putRecord`. Makes sense to throw `StreamPutException` if data ends up not writing to the stream, but what if the response itself errored out? or is that not possible? If there is no data returned (or an error), then we don't want to even attempt to write it to the stream.

model-engine/model_engine_server/inference/async_inference/tasks.py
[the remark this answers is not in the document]
tiffzhao5: seems like this doesn't work, I can't import from `api` in `inference`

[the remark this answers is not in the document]
seanshi-scale: wait is this where the async forwarder is implemented? if not, I think this code might be dead code (i.e. it was used back when async task receive/push-result would run in the same container as the user code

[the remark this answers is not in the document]
seanshi-scale: if the code is dead code, then I think the import would work because there was at some point some amount of copying only part of the code to the user inference container because of security reasons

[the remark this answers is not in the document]
tiffzhao5: yeah this is dead code so the import works but I'm referring to everywhere else (like in forwarding.py) it wouldn't work. also, if this is dead code anyways, then there's not really a point in importing from external interfaces?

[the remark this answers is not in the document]
seanshi-scale: IMO we could probably just delete the dead code (as long as we're sure it's dead)

re forwarding.py I think it should be possible to make ExternalInterfaces importable; I guess since we have stuff in `inference/{domain|infra}/gateways` the current layout of things is a bit different; I still think that in the ideal state we'd be able to use the same plugins mechanism to control which interfaces we're using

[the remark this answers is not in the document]
tiffzhao5: hm how do we make it possible to be importable in forwarding.py? yeah ideally it does make sense to control which interfaces we're using.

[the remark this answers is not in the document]
tiffzhao5: I think for now I'll leave the 2 gateways like this and can always follow-up with deleting dead code / using ExternalInterfaces

PR #433, opened 2024-02-07
Allow pydantic 2 in python client requested requirements

seanshi-scale:
# Pull Request Summary

Previous pr allowed pydantic 2 to be compatible with the client, however we didn't update the requirements to note this.

## Test Plan and Usage Guide

Pip install -e and check which version of pydantic it asks for. It asks for `pydantic >= 1.10` which is correct.

### DIFF
--- a/clients/python/llmengine/data_types.py
+++ b/clients/python/llmengine/data_types.py
@@ -10,7 +10,7 @@ import pydantic
 if int(pydantic.__version__.split(".")[0]) > 1:
     from pydantic.v1 import BaseModel, Field, HttpUrl
 else:
-    from pydantic import BaseModel, Field, HttpUrl
+    from pydantic import BaseModel, Field, HttpUrl  # type: ignore
 
 CpuSpecificationType = Union[str, int, float]
 StorageSpecificationType = Union[str, int, float]  # TODO(phil): we can make this more specific.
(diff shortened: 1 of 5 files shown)


clients/python/llmengine/data_types.py
```diff
@@ -163,17 +163,17 @@ class CreateLLMEndpointRequest(BaseModel):
     cpus: CpuSpecificationType
     gpus: int
     memory: StorageSpecificationType
-    gpu_type: GpuType
+    gpu_type: Optional[GpuType]
```
ian-scale, line 166: just curious, why not default GpuType to `None` as well?

ian-scale (approved):

clients/python/llmengine/data_types.py
yunfeng-scale, line 166: hmm, why is this optional, will we deploy LLMs on CPU?

seanshi-scale: The location in the code does specify a value for GpuType, and I think it's pretty reasonable to make us specify some value of GpuType. For most of the other ones I changed, there isn't a value passed in, so None seems like a reasonable default.

seanshi-scale: offline: we're gonna add some defaults for the model architectures, I'm gonna keep this as optional since we may want overriding of defaults at some point

seanshi-scale: for posterity: the `Optional` comes from the argument to `Model.create()`

PR #439, opened 2024-02-20
Clean up logs for logging hook

tiffzhao5:
# Pull Request Summary

Clean up logs for logging hook and add better error info.

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
+++ b/model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
@@ -8,7 +8,7 @@ class StreamingStorageGateway(ABC):
     """
 
     @abstractmethod
-    def put_record(self, stream_name: str, record: Dict[str, Any]) -> None:
+    def put_record(self, stream_name: str, record: Dict[str, Any]) -> Any:
         """
         Put a record into a streaming storage mechanism.
 
--- a/model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
+++ b/model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
@@ -8,7 +8,7 @@ class StreamingStorageGateway(ABC):
     """
 
     @abstractmethod
-    def put_record(self, stream_name: str, record: Dict[str, Any]) -> Any:
+    def put_record(self, stream_name: str, record: Dict[str, Any]) -> Dict[str, Any]:
         """
         Put a record into a streaming storage mechanism.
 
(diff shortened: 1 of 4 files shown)


model-engine/model_engine_server/inference/post_inference_hooks.py
```diff
@@ -135,9 +135,13 @@ def handle(
             logger.warning("No firehose stream name specified. Logging hook will not be executed.")
             return
         try:
-            self._streaming_storage_gateway.put_record(stream_name=stream_name, record=data_record)
-        except StreamPutException as e:
-            logger.error(f"Error in logging hook {e}")
+            firehose_response = self._streaming_storage_gateway.put_record(
```
seanshi-scale, line 138: nit: should we call this response something not specifically firehose-related? since we're in a file that shouldn't know about the streaming storage gateway implementation details

something like `streaming_storage_response`

model-engine/model_engine_server/inference/domain/gateways/streaming_storage_gateway.py
```diff
@@ -8,7 +8,7 @@ class StreamingStorageGateway(ABC):
     """
 
     @abstractmethod
-    def put_record(self, stream_name: str, record: Dict[str, Any]) -> None:
+    def put_record(self, stream_name: str, record: Dict[str, Any]) -> Any:
```
seanshi-scale, line 11: nit: it looks like we're implicitly assuming this returns a Dict with a key `ResponseMetadata` where we use the return value, looks like `TypedDict` doesn't quite work since there might be extra values returned, could we have this return a Dict type (or otherwise constrain the return type?)

model-engine/model_engine_server/inference/infra/gateways/firehose_streaming_storage_gateway.py
```diff
@@ -43,7 +43,7 @@ def _get_firehose_client(self):
         firehose_client = session.client("firehose", region_name=infra_config().default_region)
         return firehose_client
 
-    def put_record(self, stream_name: str, record: Dict[str, Any]) -> None:
+    def put_record(self, stream_name: str, record: Dict[str, Any]) -> Any:
```
seanshi-scale, line 46: same as above

seanshi-scale (commented):
had some comments about typing/information hiding, other than that it looks good

seanshi-scale (approved):

model-engine/model_engine_server/inference/post_inference_hooks.py
```diff
@@ -135,9 +135,13 @@ def handle(
...
         try:
-            self._streaming_storage_gateway.put_record(stream_name=stream_name, record=data_record)
-        except StreamPutException as e:
-            logger.error(f"Error in logging hook {e}")
+            streaming_storage_response = self._streaming_storage_gateway.put_record(
+                stream_name=stream_name, record=data_record
+            )
+        except StreamPutException:
+            logger.error(
+                f"Failed to put record into firehose stream {stream_name}. Response metadata {streaming_storage_response['ResponseMetadata']}."
```
saiatmakuri, line 143: change to .get() in case error is thrown in the fn. also set as dict?

PR #450, opened 2024-02-23
Add emitted timestamp for logging

tiffzhao5:
# Pull Request Summary

Logs emitted at timestamp for logging. 

## Test Plan and Usage Guide
image

### DIFF
--- a/model-engine/model_engine_server/inference/post_inference_hooks.py
+++ b/model-engine/model_engine_server/inference/post_inference_hooks.py
@@ -124,7 +124,7 @@ class LoggingHook(PostInferenceHook):
             return
         response["task_id"] = task_id
         data_record = {
-            "EMITTED_AT": datetime.now(pytz.timezone("US/Pacific")).strftime("%Y-%m-%dT%H:%M:%S"),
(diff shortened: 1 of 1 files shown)


model-engine/model_engine_server/inference/post_inference_hooks.py
```diff
@@ -122,6 +124,7 @@ def handle(
             return
         response["task_id"] = task_id
         data_record = {
+            "EMITTED_AT": datetime.now(pytz.timezone("US/Pacific")).strftime("%Y-%m-%dT%H:%M:%S"),
```
saiatmakuri, line 127: nit: should be a UTC timestamp so there's no timezone or daylight saving time information baked into the dates here

saiatmakuri (approved):

model-engine/model_engine_server/inference/post_inference_hooks.py
tiffzhao5: i'm fine with using UTC but seems like we're using PST elsewhere in the codebase? 
https://github.com/scaleapi/llm-engine/blob/31c7c5a392a1265c52b7f70e8e08d20ca293f697/model-engine/model_engine_server/api/llms_v1.py#L125

PR #465, opened 2024-03-08
Change back batch infer GPU util and add tool completion client changes

yunfeng-scale:
# Pull Request Summary

1. 80% mem usage is probably too conservative. just print mem usage for now
2. add tool completion client changes

## Test Plan and Usage Guide

already deployed

### DIFF
--- a/clients/python/setup.py
+++ b/clients/python/setup.py
@@ -3,6 +3,6 @@ from setuptools import find_packages, setup
 setup(
     name="scale-llm-engine",
     python_requires=">=3.7",
-    version="0.0.0.beta25",
+    version="0.0.0.beta26",
     packages=find_packages(),
 )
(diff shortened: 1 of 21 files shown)


seanshi-scale (approved):

saiatmakuri (approved):

clients/python/setup.py
```diff
@@ -3,6 +3,6 @@
 setup(
     name="scale-llm-engine",
     python_requires=">=3.7",
-    version="0.0.0.beta25",
+    version="0.0.0.beta26",
```
seanshi-scale, line 6: could we make the version bump change in pyproject.toml also?

seanshi-scale (commented):
note about bumping the version in the third of three places (pyproject.toml), rest lgtm

clients/python/setup.py
yunfeng-scale: ah thanks for the catch

seanshi-scale (approved):

PR #466, opened 2024-03-09
Try to fix async requests getting stuck

squeakymouse:
# Pull Request Summary

Change SQS broker options for Celery; decrease wait time for SQS long polling

Not strictly necessary, but also changed `echo_server` for testing

## Test Plan and Usage Guide

Test deployment

### DIFF
--- a/clients/python/llmengine/__init__.py
+++ b/clients/python/llmengine/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.0b25"
+__version__ = "0.0.0b26"
 
 import os
 from typing import Sequence
--- a/clients/python/llmengine/completion.py
+++ b/clients/python/llmengine/completion.py
@@ -10,6 +10,7 @@ from llmengine.data_types import (
     CreateBatchCompletionsRequest,
     CreateBatchCompletionsRequestContent,
     CreateBatchCompletionsResponse,
+    ToolConfig,
 )
 
 COMPLETION_TIMEOUT = 300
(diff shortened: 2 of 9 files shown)


model-engine/model_engine_server/core/celery/app.py
```diff
@@ -475,6 +475,8 @@ def _get_broker_endpoint_and_transport_options(
         # backoff_policy, etc., then we can expose broker_transport_options in the top-level celery() wrapper function.
         # Going to try this with defaults first.
         out_broker_transport_options["region"] = os.environ.get("AWS_REGION", "us-west-2")
+        out_broker_transport_options["wait_time_seconds"] = 0
```
ian-scale, line 478: just curious: can you explain how this works?

model-engine/model_engine_server/core/celery/app.py
```diff
@@ -475,6 +475,8 @@ def _get_broker_endpoint_and_transport_options(
...
+        out_broker_transport_options["polling_interval"] = 5
```
ian-scale, line 479: also does this add any additional overhead for async requests because we're polling more often? would the volume of async GET requests scale proportionally to this

model-engine/model_engine_server/core/celery/app.py
seanshi-scale: +1, would be good to explain what's going on with these changes, and/or any testing results showing the impact of the changes.

model-engine/model_engine_server/core/celery/app.py
seanshi-scale: I think this controls how often the celery workers poll SQS. It shouldn't affect any quantity of inbound requests to the gateway.

seanshi-scale (approved):
Could you also put the explanation for `wait_time_seconds=0` in the code as well? Think it's something that's quite nontrivial and would make it a lot less confusing in the future.

PR #468, opened 2024-03-12
Azure fixes + additional asks

squeakymouse:
# Pull Request Summary

## Test Plan and Usage Guide


model-engine/model_engine_server/db/base.py
```diff
@@ -81,7 +81,9 @@ def get_engine_url(env: Optional[str] = None, read_only: bool = True, sync: bool
 
     # For async postgres, we need to use an async dialect.
     if not sync:
-        engine_url = engine_url.replace("postgresql://", "postgresql+asyncpg://")
+        engine_url = engine_url.replace("postgresql://", "postgresql+asyncpg://").replace(
```
seanshi-scale, line 84: what does `replace('sslmode', 'ssl')` do? does it remain compatible with our aws db setup?

model-engine/model_engine_server/common/config.py
```diff
@@ -81,9 +83,17 @@ def cache_redis_url(self) -> str:
...
         username = os.getenv("AZURE_OBJECT_ID")
-        password = DefaultAzureCredential().get_token("https://redis.azure.com/.default").token
+        token = DefaultAzureCredential().get_token("https://redis.azure.com/.default")
+        password = token.token
+        global redis_cache_expiration
+        redis_cache_expiration = token.expires_on
         return f"rediss://{username}:{password}@{self.cache_redis_azure_host}"
 
+    @property
+    def cache_redis_url_expiration(self) -> Optional[int]:
```
seanshi-scale, line 93: I originally thought this was something representing a "timedelta", should we call it `cache_redis_url_expiration_timestamp`? (or otherwise make it clear it represents a timestamp/point in time)

model-engine/model_engine_server/infra/services/image_cache_service.py
```diff
@@ -78,11 +79,14 @@ def _cache_finetune_llm_images(
...
-            if not CIRCLECI
-            else "fake_docker_repository_latest_image_tag"
-        )
+        latest_tag = "fake_docker_repository_latest_image_tag"
+        if not CIRCLECI:
+            try:
+                latest_tag = self.docker_repository.get_latest_image_tag(
+                    hmi_config.batch_inference_vllm_repository
+                )
+            except ResourceNotFoundError:
```
seanshi-scale, line 88: If we wanted to be strict about clean architecture, I feel like we'd probably want to have the service not need to know about azure (which means having the docker repository bubble up an error that was not specific to AWS/Azure/etc., and have the docker repositories themselves catch the cloud-specific error and raise the non-cloud-specific error).

Also, would we want to catch any errors that AWS/boto3 might throw at us as well?

model-engine/model_engine_server/api/files_v1.py
```diff
@@ -44,7 +44,7 @@ async def upload_file(
     )
     return await use_case.execute(
         user=auth,
-        filename=file.filename,
+        filename=file.filename or "",
```
seanshi-scale, line 47: when does `file.filename` equal None, should we just throw a 4xx in that case (if it is user fault that `file.filename` could be None?) 

maybe this is fine if `filename` doesn't get used as the only part of an identifier actually, tbh I'm not sure though

model-engine/model_engine_server/infra/repositories/acr_docker_repository.py
```diff
@@ -36,7 +37,10 @@ def get_latest_image_tag(self, repository_name: str) -> str:
...
 
-        image = client.list_manifest_properties(
-            repository_name, order_by="time_desc", results_per_page=1
-        ).next()
-        return image.tags[0]
+        try:
+            image = client.list_manifest_properties(
+                repository_name, order_by="time_desc", results_per_page=1
+            ).next()
+            return image.tags[0]
```
seanshi-scale, line 44: do we want to throw an error if there are 0 image tags?

seanshi-scale: unless that's already handled in the `ResourceNotFoundError`

seanshi-scale (approved):
had a few more comments, but they're pretty minor. LGTM once they're addressed

model-engine/model_engine_server/api/files_v1.py
seanshi-scale: do we know when filename can be None? eg does the fastapi documentation say anything about it

model-engine/model_engine_server/infra/repositories/acr_docker_repository.py
seanshi-scale: ah ok sounds good, could we note it in the code so we know why we're not gonna `IndexError`?

yunfeng-scale (approved):

PR #469, opened 2024-03-12
Tool completion respect num new tokens

yunfeng-scale:
# Pull Request Summary

respect num new tokens

## Test Plan and Usage Guide

pending testing on real data

### DIFF
--- a/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
+++ b/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
@@ -260,7 +260,10 @@ async def generate_with_tool(
             gen_item.generated_text += new_text
 
             # If we didn't just execute a tool, we're done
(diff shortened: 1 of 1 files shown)


model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -264,6 +264,10 @@ def tool_func(text: str, past_context: Optional[str]):
                 gen_item.completed = True
                 continue
 
+            if gen_item.remaining_tokens <= 0:
```
Georgepu1, line 267: nit: we can just check for this in the prev conditional

Georgepu1 (approved):

ian-scale (approved):

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yunfeng-scale: sure

PR #470, opened 2024-03-13
Parallel Python execution for tool completion

yunfeng-scale:
# Pull Request Summary

with some sample data (1000 prompts). on my devbox (96 CPU cores), but since i used threadpool to start subprocess i think it still might be slowed down by GIL to start processes, probably not high util for the 96 cores

467s total, 283s tool use
vs
without this change 862s total, 687s tool use

## Test Plan and Usage Guide

tested with sample data
also CPU count works inside and outside container


model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -265,7 +270,9 @@ def tool_func(text: str, past_context: Optional[str]):
                 or gen_item.remaining_tokens <= 0
             ):
                 gen_item.completed = True
-                continue
+
+        pool = ThreadPool(CPU_COUNT)
```
yunfeng-scale, line 274: it's probably fine to use single process to start subprocesses when CPU number is low

Georgepu1 (approved):
lgtm, thanks for investigating! btw were u able to validate the completions were close enough on the "467s total, 283s tool use" vs "862s total, 687s tool use" runs? just want to make sure nothing weird's happening here on the generation side

yunfeng-scale:
took a bit more look into this @Georgepu1 15/1000 output texts have changed, i manually checked 5 of them and i don't think python code execution results are different. the differences appear to be some randomness in sampling that sentences in both cases made sense to me

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -30,6 +32,8 @@
 
 os.environ["AWS_PROFILE"] = os.getenv("S3_WRITE_AWS_PROFILE", "default")
 
+CPU_COUNT = multiprocessing.cpu_count()
```
yixu34, line 35: I remember that there may be an issue with this, where this returns the # of CPUs on the host, as opposed the # of CPUs allotted to the pod in k8s.

yunfeng-scale: good point. let me dig a bit

yunfeng-scale: use more accurate calculations

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -31,6 +33,25 @@
 os.environ["AWS_PROFILE"] = os.getenv("S3_WRITE_AWS_PROFILE", "default")
 
 
+def get_CPU_cores_in_container():
```
yixu34, line 36: nit: `get_cpu_cores_in_container`?

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -31,6 +33,25 @@
...
+        if cfs_quota_us != -1:
+            cpu_count = cfs_quota_us // cfs_period_us
+    except FileNotFoundError:
+        pass
+    return cpu_count
+
+
+CPU_COUNT = get_CPU_cores_in_container()
+
+print(f"{CPU_COUNT=}")
```
yixu34, line 52: Remove?

yixu34 (approved):

PR #475, opened 2024-03-20
fix docs tests gateway endpoint

saiatmakuri:
# Pull Request Summary

`integration_tests/test_docs` pointed to default gateway instead of integration test gateway. fix this

## Test Plan and Usage Guide

`just _integration_test training test_docs`

### DIFF
--- a/integration_tests/test_docs.py
+++ b/integration_tests/test_docs.py
@@ -9,7 +9,12 @@ from textwrap import dedent
 import pytest
 from _pytest.assertion.rewrite import AssertionRewritingHook
 
-from .rest_api_utils import BASE_PATH, SERVICE_IDENTIFIER
+from .rest_api_utils import (
+    BASE_PATH,
+    SERVICE_IDENTIFIER,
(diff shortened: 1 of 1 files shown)


ian-scale (approved):

integration_tests/test_docs.py
```diff
@@ -56,11 +57,17 @@ def integration_test_user_id() -> str:
 
 
 def modify_source(source: str) -> str:
-    # Adds some custom logic to update code from docs to comply with some requirements.
-    source = re.sub(r"('team'|\"team\"): ('\w+'|\"\w+\")", r"'team': 'infra'", source)
+    """Adds some custom logic to update code from docs to comply with some requirements."""
```
ian-scale, line 60: nit: docstring is super vague, could we clarify

ian-scale (approved):

integration_tests/test_docs.py
saiatmakuri: updated comments in the fn to explain actions

PR #478, opened 2024-03-28
Downgrade sse-starlette version

squeakymouse:
# Pull Request Summary

The updated version has weird behavior with streaming in the http-forwarder (tokens getting streamed back were batched so time to first token was very long; we expect to get tokens back at a steady rate instead)

I tried sse-starlette version 1.8.2 to see if it was the major version bump (1.8.2 -> 2.0.0) that broke things, but 1.8.2 still had the weird behavior, so downgrading to the original version of 1.6.1 from before the security scan updates

Should still be fine with security scan

## Test Plan and Usage Guide

Tested that using this image for the HTTP forwarder of a Llama 2 endpoint in the training cluster fixes the oncall issue of streaming time to first token being long (via curling localhost:5000 from the HTTP forwarder)

### DIFF
--- a/model-engine/requirements.in
+++ b/model-engine/requirements.in
@@ -47,7 +47,7 @@ sentencepiece==0.1.99
 sh~=1.13
 smart-open~=5.2
 sqlalchemy[asyncio]~=2.0.4
-sse-starlette==1.8.2
+sse-starlette==1.6.1
 sseclient-py==1.7.2
 starlette[full]>=0.36.2 # not used directly, but needs to be pinned for Microsoft security scan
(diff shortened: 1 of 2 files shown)


ian-scale:
> Should still be fine with Microsoft security scan

any way to confirm this for sure before merging?

seanshi-scale (commented):
could you write how changes were tested?

edgan8 (approved):
Talked with Katie, sse starlette was on an older version two weeks ago so this looks good. Could you add more details to the PR explaining the context for what broke and why we need this specific version?

PR #479, opened 2024-03-30
Return 400 for botocore client errors

yunfeng-scale:
# Pull Request Summary

Return 400 for some client errors

## Test Plan and Usage Guide

tested calling an endpoint with long message and received:
`<Response [400]>`
```
{"detail":"Invalid request: Error sending celery task: An error occurred (InvalidParameterValue) when calling the SendMessage operation: One or more parameters are invalid. Reason: Message must be shorter than 262144 bytes."}
```

### DIFF
--- a/model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
+++ b/model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
@@ -1,5 +1,6 @@
 from typing import Any, Dict, List, Optional
 
+import botocore
 from model_engine_server.common.dtos.model_endpoints import BrokerType
 from model_engine_server.common.dtos.tasks import (
     CreateAsyncTaskV1Response,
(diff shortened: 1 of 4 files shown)


model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
```diff
@@ -68,12 +70,16 @@ def send_task(
...
-            queue=queue_name,
-        )
+        try:
+            res = celery_dest.send_task(
+                name=task_name,
+                args=args,
+                kwargs=kwargs,
+                queue=queue_name,
+            )
+        except botocore.exceptions.ClientError as e:
```
edgan8, line 80: Is there a more specific error we can catch so we don't inadvertently classify internal bugs as 400s?

edgan8 (approved):

model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
yunfeng-scale: yes i can check the error message whether it's about message size. but for now i want to make it simple and count every boto client error as user errors :) we can revisit it later when users complain.

PR #480, opened 2024-03-30
Batch job metrics

yunfeng-scale:
# Pull Request Summary

add some metrics to batch jobs

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
+++ b/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
@@ -10,7 +10,6 @@ from urllib.parse import urlparse
 
 import boto3
 import smart_open
-from datadog import statsd
 from func_timeout import FunctionTimedOut, func_set_timeout
 from model_engine_server.common.dtos.llms import (
     CompletionOutput,
--- a/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
+++ b/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
@@ -420,10 +420,10 @@ async def generate_with_vllm(
         if return_token_log_probs:
             output.tokens = tokens
 
-        tags = {
-            "model": model,
-            "use_tool": use_tool,
-        }
+        tags = [
+            f"model:{model}",
+            f"use_tool:{use_tool}",
+        ]
         statsd.increment(
             "model_engine.batch_inference.vllm.generation_count",
             tags=tags,
--- a/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
+++ b/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
@@ -148,6 +148,7 @@ async def generate_with_tool(
     content: CreateBatchCompletionsRequestContent,
     prompts,
     tool: Type[BaseTool],
+    is_finetuned: bool,
 ):
     class IterativeGeneration:
         def __init__(self, prompt, max_new_tokens):
@@ -199,6 +200,7 @@ async def generate_with_tool(
             [iter[0] for iter in iter_prompts],
             bar,
             use_tool=True,
+            is_finetuned=is_finetuned,
         )
 
         bar = tqdm(
@@ -300,6 +302,7 @@ async def batch_inference():
     model = (
         MODEL_WEIGHTS_FOLDER if request.model_config.checkpoint_path else request.model_config.model
     )
+    is_finetuned = request.model_config.checkpoint_path is not None
 
     llm = get_vllm_engine(model, request)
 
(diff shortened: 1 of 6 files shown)


model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -414,6 +420,30 @@ async def generate_with_vllm(
         if return_token_log_probs:
             output.tokens = tokens
 
+        tags = [
+            f"model:{model}",
```
saiatmakuri, line 424: is this the base model name or finetuned model?

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -414,6 +420,30 @@ async def generate_with_vllm(
...
+            f"use_tool:{use_tool}",
+        ]
+        statsd.increment(
+            "model_engine.batch_inference.vllm.generation_count",
```
saiatmakuri, line 428: use [metrics gateway](https://github.com/scaleapi/llm-engine/blob/main/model-engine/model_engine_server/inference/infra/gateways/datadog_inference_monitoring_metrics_gateway.py#L7)?

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -414,6 +420,30 @@ async def generate_with_vllm(
...
+            tags=tags,
+        )
+        statsd.increment(
+            "model_engine.batch_inference.vllm.token_count.total",
+            num_prompt_tokens + num_completion_tokens,
```
edgan8, line 433: Is this possible to just compute when querying metrics later?

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -414,6 +420,30 @@ async def generate_with_vllm(
         if return_token_log_probs:
             output.tokens = tokens
 
+        tags = [
+            f"model:{model}",
+            f"use_tool:{use_tool}",
```
edgan8, line 425: Is use_tool actually used to control any behavior? Seems kind of funny to have a parameter that is not linked to any behavioral change.

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
seanshi-scale: I think if we do increment we can just add them later, but if we want to do something more fancy like histogram/distribution then we need to output the total separately

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yunfeng-scale: it's used to control whether we want to run python code

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yunfeng-scale: base model. there's no name for the finetuned model. i can add a tag for the checkpoint path, but that'll be a very long string and not sure if useful?

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yunfeng-scale: yes it is possible, also things like Sean said

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yunfeng-scale: sure

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -361,9 +367,14 @@ async def generate_with_vllm(
     top_p,
     prompts,
     bar,
+    use_tool,
 ) -> List[CompletionOutput]:  # pragma: no cover
     from vllm import SamplingParams
 
+    model = (await engine.get_model_config()).model
```
edgan8, line 374: Does this have a risk of increasing the latency?

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
```diff
@@ -10,6 +10,7 @@
 
 import boto3
 import smart_open
+from datadog import statsd
```
edgan8, line 13: This import looks like its no longer required, can we remove the datadog dependency in the requirements.txt as well?

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yunfeng-scale: it'll ask Ray workers to get model config (for some reason) and it should be pretty fast. also this call runs once for the whole batch instead of each prompt

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yunfeng-scale: yeah that's probably also causing lint error

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yixu34: We do have an `is_finetuned` flag in our metrics. Would it be useful to propagate that here too?

edgan8 (approved):

model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
yunfeng-scale: sure

PR #482, opened 2024-04-02
Change LLM Engine base path from global var

squeakymouse:
# Pull Request Summary

Needed for SGP code that uses LLM Engine

## Test Plan and Usage Guide

Tested with `pip install -e .`

### DIFF
--- a/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
+++ b/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
@@ -149,6 +149,7 @@ async def generate_with_tool(
     prompts,
     tool: Type[BaseTool],
     is_finetuned: bool,
+    model: str,
 ):
     class IterativeGeneration:
         def __init__(self, prompt, max_new_tokens):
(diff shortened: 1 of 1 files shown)


clients/python/llmengine/api_engine.py
```diff
@@ -12,12 +12,15 @@
...
 
 SPELLBOOK_API_URL = "https://api.spellbook.scale.com/llm-engine/"
-LLM_ENGINE_BASE_PATH = os.getenv("LLM_ENGINE_BASE_PATH", SPELLBOOK_API_URL)
 DEFAULT_TIMEOUT: int = 10
 
 api_key = None
 
 
+def get_base_path() -> str:
+    return os.getenv("LLM_ENGINE_BASE_PATH", SPELLBOOK_API_URL)
```
ian-scale, line 21: nit: since we're changing things with this var, should we rename SPELLBOOK as well?

saiatmakuri (approved):

seanshi-scale (approved):

PR #487, opened 2024-04-10
Improve TensorRT-LLM Functionality

seanshi-scale:
# Pull Request Summary

Changes to get tensorrtllm to work with Mixtral

1. Update tensorrt llm included code/build processes to a newer version
2. Add some bits to mitigate some tokenization issues

Note: the logprobs returned aren't correct still, haven't investigated.

## Test Plan and Usage Guide

Deployed a weights-only quantized Mixtral model, which works


model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/postprocessing/1/model.py
```diff
@@ -145,12 +181,21 @@ def finalize(self):
...
-            for tokens in beam_tokens:
-                output = self.tokenizer.decode(tokens)
-                outputs.append(output.encode("utf8"))
+        for batch_idx, beam_tokens in enumerate(tokens_batch):
+            for beam_idx, tokens in enumerate(beam_tokens):
+                seq_len = sequence_lengths[batch_idx][beam_idx]
+                output = self.tokenizer.decode(
+                    tokens[:seq_len],
+                    skip_special_tokens=self.skip_special_tokens)
+                # Adapted from https://github.com/triton-inference-server/tensorrtllm_backend/pull/423
```
seanshi-scale, line 194: Differs from NVIDIA here

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/preprocessing/1/model.py
```diff
@@ -166,59 +228,116 @@ def finalize(self):
...
+                    word = word.decode()
 
+                ids = self.tokenizer.encode(word, add_special_tokens=False)
                 if len(ids) == 0:
                     continue
 
                 item_flat_ids += ids
                 item_offsets.append(len(ids))
 
+                # Add a case where ids[0] decodes to empty string, then add another set of ids here
```
seanshi-scale, line 287: Differs from NVIDIA here

model-engine/tests/unit/domain/test_llm_use_cases.py
```diff
@@ -836,6 +836,46 @@ async def test_completion_sync_trt_llm_use_case_success(
...
+    fake_tokenizer_repository,
+    llm_model_endpoint_trt_llm: ModelEndpoint,
+    completion_sync_request: CompletionSyncV1Request,
+):
+    completion_sync_request.return_token_log_probs = False  # not yet supported
+    fake_llm_model_endpoint_service.add_model_endpoint(llm_model_endpoint_trt_llm)
+    fake_model_endpoint_service.sync_model_endpoint_inference_gateway.response = SyncEndpointPredictV1Response(
+        status=TaskStatus.SUCCESS,
+        result={
+            "result": '{"context_logits":0.0,"cum_log_probs":0.0,"generation_logits":0.0,"model_name":"ensemble","model_version":"1","output_log_probs":[0.0,0.0,0.0,0.0,0.0],"sequence_end":false,"sequence_id":0,"sequence_start":false,"text_output":" Machine learning is a branch"}'
```
seanshi-scale, line 857: may need to figure out why the log probs are not returned properly

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/postprocessing/1/model.py
```diff
@@ -147,10 +176,25 @@ def finalize(self):
...
+    def _postprocessing(self, tokens_batch, sequence_lengths):
         outputs = []
-        for beam_tokens in tokens_batch:
-            for tokens in beam_tokens:
-                output = self.tokenizer.decode(tokens)
+        for batch_idx, beam_tokens in enumerate(tokens_batch):
+            for beam_idx, tokens in enumerate(beam_tokens):
+                seq_len = sequence_lengths[batch_idx][beam_idx]
+                output = self.tokenizer.decode(
+                    tokens[:seq_len], skip_special_tokens=self.skip_special_tokens
```
yunfeng-scale, line 185: why do we restrict to [:seq_len], what are in `tokens` that outside of `seq_len`?

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/postprocessing/1/model.py
```diff
@@ -147,10 +176,25 @@ def finalize(self):
...
+                )
+                # Adapted from https://github.com/triton-inference-server/tensorrtllm_backend/pull/423
+                # This is somewhat of a hack: add a space before the output if the first token starts with a space
+                # This may add a space in front of the first token though when we don't want it.
+                token_id_string = self.tokenizer.convert_ids_to_tokens(
```
yunfeng-scale, line 190: this is expensive though? we're effectively decoding twice. if we're only checking string for first token, can we just do ```
self.tokenizer.convert_ids_to_tokens(
                    tokens[0], skip_special_tokens=self.skip_special_tokens
                )```?

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/preprocessing/1/model.py
yunfeng-scale: why do we do this?

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/postprocessing/1/model.py
seanshi-scale: It's present in the original code https://github.com/triton-inference-server/tensorrtllm_backend/blob/v0.8.0/all_models/inflight_batcher_llm/postprocessing/1/model.py#L192

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/preprocessing/1/model.py
seanshi-scale: This is a partial patch to get some of the stop sequence behavior more functional; when I was trying stop sequences out, I noticed that there were cases where the stop sequence was being ignored. Root cause was because trt tokenizes the stop sequences, and looks for that exact sequence of tokens; but there was a case where what's returned from the model isn't quite that stop sequence, i.e. there's an extra empty token when trt tokenizes the stop sequence. So I just patched it so that we also look for the original sequence minus the empty token.

ex. if you pass in a stop sequence of `text` that tokenizes to `[1,2,3]`, the model can output the sequence `[..., 2,3]`, where `[2,3]` also decodes to `text`.

yunfeng-scale: even this doesn't seem right? stop sequence could be a part of [2,3], e.g. tok2 = abc, tok3 = def, stop sequence = cdef. i think best is to compare in postprocessing with string?

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/postprocessing/1/model.py
```diff
@@ -147,10 +176,26 @@ def finalize(self):
...
-    def _postprocessing(self, tokens_batch):
+    def _postprocessing(self, tokens_batch, sequence_lengths):
         outputs = []
-        for beam_tokens in tokens_batch:
-            for tokens in beam_tokens:
-                output = self.tokenizer.decode(tokens)
+        for batch_idx, beam_tokens in enumerate(tokens_batch):
+            for beam_idx, tokens in enumerate(beam_tokens):
+                seq_len = sequence_lengths[batch_idx][beam_idx]
+                output = self.tokenizer.decode(
```
yunfeng-scale, line 184: should we check for stop token here?

yunfeng-scale (approved):

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/preprocessing/1/model.py
seanshi-scale: yup, this isn't a complete fix for stop sequences for sure unfortunately; think we'd need to spend more time to see if it's possible with the current framework and how to do it

model-engine/model_engine_server/inference/tensorrt-llm/triton_model_repo/postprocessing/1/model.py
seanshi-scale: I don't remember seeing stop tokens in the output when testing at least.

PR #493, opened 2024-04-16
Properly add mixtral 8x22b

yunfeng-scale:
# Pull Request Summary

## Test Plan and Usage Guide

attempted to create such an endpoint locally

### DIFF
--- a/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
@@ -180,6 +180,7 @@ _SUPPORTED_MODELS_BY_FRAMEWORK = {
             "mistral-7b-instruct",
             "mixtral-8x7b",
             "mixtral-8x7b-instruct",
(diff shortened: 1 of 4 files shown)


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -180,6 +180,7 @@
             "mistral-7b-instruct",
             "mixtral-8x7b",
             "mixtral-8x7b-instruct",
+            "mixtral-8x22b",
```
saiatmakuri, line 183: is there an instruct version to add too?

saiatmakuri (approved):

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: don't see an instruct version from huggingface

yixu34 (approved):

PR #498, opened 2024-04-23
fix return_token_log_probs on vLLM > 0.3.3 endpoints

saiatmakuri:
# Pull Request Summary

since vLLM > 0.3.3, `logprobs` are returned as as type `Logprob` instead as direct float's. it's an API-breaking change (see https://github.com/vllm-project/vllm/pull/3065#issuecomment-1984761326)

change `vllm_server` to return output in the same previous format

## Test Plan and Usage Guide

create new `llama-2-7b` with vLLM==0.4.0.post1

```python
Completion.create(
    model="llama-2-7b",
    prompt="Hello ",
    max_new_tokens=10,
    temperature=0.0,
    return_token_log_probs=True,
)
```

### DIFF
--- a/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
+++ b/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
@@ -409,11 +409,6 @@ async def generate_with_vllm(
 ) -> List[CompletionOutput]:  # pragma: no cover
     from vllm import SamplingParams
 
-    try:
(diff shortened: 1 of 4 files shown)


yunfeng-scale (approved):
thanks for the fix! i think we need to do some backfills also?

saiatmakuri:
> thanks for the fix! i think we need to do some backfills also?

will update endpoints to latest image post merge

PR #505, opened 2024-05-02
get latest inference framework tag from configmap

saiatmakuri:
# Pull Request Summary

use a remote configmap to manage latest inference framework versions

## Test Plan and Usage Guide

try create request with `"latest"` tag with "vllm" and "tensortrt_llm" frameworks


charts/model-engine/templates/inference_framework_config.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: model-engine-inference-framework-latest-config
+  labels:
+    product: common
+    team: infra
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
```
seanshi-scale, line 9: should this be just `pre-install`? Yunfeng told me the idea behind this is to decouple upgrading inference framework versions and installs of model-engine, and if we do `pre-upgrade` I think that updates on installing a new version of model-engine

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -246,6 +248,8 @@
 NUM_DOWNSTREAM_REQUEST_RETRIES = 80  # has to be high enough so that the retries take the 5 minutes
 DOWNSTREAM_REQUEST_TIMEOUT_SECONDS = 5 * 60  # 5 minutes
 
+LATEST_CONFIG_MAP_NAME = "model-engine-inference-framework-latest-config"
```
seanshi-scale, line 251: nit: `LATEST_INFERENCE_FRAMEWORK...`? feels a bit ambiguous as it is currently heh

charts/model-engine/templates/inference_framework_config.yaml
seanshi-scale: also would it be possible to mention why we're putting this in a new configmap?

charts/model-engine/templates/inference_framework_config.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: model-engine-inference-framework-latest-config
```
yunfeng-scale, line 4: shouldn't this configmap be in the same namespace with all the other stuff?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -931,8 +944,8 @@ async def execute(
                 )
 
         if request.inference_framework_image_tag == "latest":
-            request.inference_framework_image_tag = self.docker_repository.get_latest_image_tag(
```
yunfeng-scale, line 934: i thought we're going to fallback to `get_latest_image_tag` when the configmap doesn't work?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -255,6 +259,15 @@ def count_tokens(input: str, model_name: str, tokenizer_repository: TokenizerRep
     return len(tokenizer.encode(input))
 
 
+async def _get_latest_tag(inference_framework: LLMInferenceFramework) -> str:
+    config_map = await read_config_map(LATEST_INFERENCE_FRAMEWORK_CONFIG_MAP_NAME)
+    if inference_framework not in config_map:
+        raise LatestImageTagNotFoundException(
+            f"Could not find latest tag for inference framework {inference_framework}."
+        )
+    return config_map[inference_framework]
```
yunfeng-scale, line 268: let's also check whether the image actually exists

charts/model-engine/templates/inference_framework_config.yaml
saiatmakuri: it should be in the same namespace as the gateway, so default

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
saiatmakuri: this happens later in the creation process

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
saiatmakuri: I think we can just error if "latest" doesn't exist

charts/model-engine/templates/inference_framework_config.yaml
seanshi-scale: with sgp installs we're not guaranteed that it's in the default namespace

yunfeng-scale: i think you can record model engine namespace in service config yaml as
```
model_engine_namespace: {{.Release.Namespace}}
```

charts/model-engine/values_circleci.yaml
```diff
@@ -96,6 +96,8 @@ config:
       profile_ml_worker: "default"
       profile_ml_inference_worker: "default"
     launch:
+      # Config to know where model-engine is running
+      gateway_namespace : default
```
seanshi-scale, line 100: ideally we'd be able to not need to specify the namespace a second time here, since we pass that through the helm install command (e.g. `helm install ... --namespace <namespace>`) and can get the value through `{{.Release.namespace}}` (or whatever it is).
Having said that, getting that working also might be a bit ugly (e.g. we'd auto inject `gateway_namespace` into the config but nothing else)

charts/model-engine/templates/service_config_map.yaml
```diff
@@ -11,6 +11,7 @@ metadata:
 data:
   launch_service_config: |-
     dd_trace_enabled: {{ .Values.dd_trace_enabled | default false | quote }}
+    gateway_namespace: {{ .Release.Namespace }}
```
seanshi-scale, line 14: do we need to pass it through that ` | quote` thing? we could helm install --dryrun to see what the service config map ends up as

seanshi-scale: (or `just _deploy training true` actually heh which is probably easier)

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: you meant within `create_llm_model_bundle_use_case.execute`?

yunfeng-scale (approved):

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
saiatmakuri: we first map "latest" to the value from the config map. then in `create_llm_model_bundle_use_case.execute` that tag is checked

PR #515, opened 2024-05-07
Infer hardware from model name

yunfeng-scale:
# Pull Request Summary

Infer hardware specs from model name so these fields are optional now

formular used:
```
dtype_size = 2 (for 16bit float)

min_kv_cache_size =
        2
        * dtype_size
        * config["num_hidden_layers"]
        * config["hidden_size"]
        * config["max_position_embeddings"]
        // (config["num_attention_heads"] // config["num_key_value_heads"])

model_weights_size = dtype_size * model_param_count_b * 1_000_000_000

min_memory_gb = math.ceil((min_kv_cache_size + model_weights_size) / 1_000_000_000 / 0.9)
```

we hard code the param count for MoE models right now.
we omit some other small weights like embedding layer and MLP layer post transformer
this estimates is mostly correct with some issues on long context window models (investigation TBD)

## Test Plan and Usage Guide

added unit tests and created llama 3 8b and codellama endpoint


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2201,6 +2164,32 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
         return ModelDownloadResponse(urls=urls)
 
 
+def fill_hardware_info(request: CreateLLMModelEndpointV1Request):
```
yixu34, line 2167: nit: should we prefix these module-level functions with an underscore? Don't want to tempt people to import those, when they should be going through the use case.

yixu34 (commented):
Do we have a use case-level test that verifies this behavior? If there was a prior test that exercised `infer_hardware_from_model_name`, we can probably just route to that one.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -219,29 +219,6 @@
     LLMInferenceFramework.TENSORRT_LLM: [],
 }
 
-# We need a dict where if we need to override we can
```
seanshi-scale, line 222: what's the reason we're getting rid of the `max_model_len`/`max_num_batched_tokens` args to vllm?

yunfeng-scale: i forgot why we're doing this in the first place, but i'm pretty certain in recent version of vLLM this is not needed. also checked most of these models, config.json has the same `max_position_embeddings` with the values here

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2198,7 +2169,33 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
...
+    ):
+        if not (
+            request.gpus is None
+            and request.gpu_type is None
+            and request.cpus is None
+            and request.memory is None
+            and request.storage is None
+        ):
+            raise ObjectHasInvalidValueException(
+                "All hardware spec fields (gpus, gpu_type, cpus, memory, storage) must be provided if any hardware spec field is provided."
```
yixu34, line 2188: I'm not sure if this sentence aligns with the logic?

```python
if not (
            request.gpus is None
            and request.gpu_type is None
            and request.cpus is None
            and request.memory is None
            and request.storage is None
        ):
```
seems to align with `All hardware spec fields [...] must be provided [...]` but

```python
 if (
        request.gpus is None
        or request.gpu_type is None
        or request.cpus is None
        or request.memory is None
        or request.storage is None
    ):
```
Seems to say "if any of these fields is missing", not "if any of these fields is provided".

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2219,7 +2216,7 @@ def infer_hardware_from_model_name(model_name: str) -> CreateDockerImageBatchJob
             )
 
         b_params = int(numbers[-1])
-        if b_params <= 7:
+        if b_params <= 8:
```
yixu34, line 2219: This magic number seems a bit brittle.

yixu34 (commented):
👀

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: best scenario i need to come up with a formula that incorporates # billions of parameters and context window length (for MoE add # of experts, # billion of parameters shared and for each expert), and output GB memory needed, from there I can calculate GPU type and sharding scheme. i feel this is too much work to justify

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: yeah good catch

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: will do a bit more calculation and less hardcode

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2200,14 +2172,106 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
...
+        request.memory = hardware_info.memory
+        request.storage = hardware_info.storage
+
+
+def _infer_hardware(
+    llm_artifact_gateway: LLMArtifactGateway,
+    model_name: str,
+    checkpoint_path: str,
+) -> CreateDockerImageBatchJobResourceRequests:
+    print("getting config")
```
yixu34, line 2209: Remove prints?

model-engine/model_engine_server/infra/gateways/abs_llm_artifact_gateway.py
```diff
@@ -73,3 +74,11 @@ def get_model_weights_urls(self, owner: str, model_name: str, **kwargs) -> List[
         for blob_name in container_client.list_blob_names(name_starts_with=prefix):
             model_files.append(f"https://{account}.blob.core.windows.net/{bucket}/{blob_name}")
         return model_files
+
+    def get_model_config(self, path: str, **kwargs) -> Dict[str, Any]:
```
yixu34, line 78: Are these changes related...? I think these are @squeakymouse ?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2200,14 +2172,106 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
...
+            raise ObjectHasInvalidValueException(
+                f"Unable to infer number of parameters for {model_name}."
+            )
+        model_param_count_b = int(numbers[-1])
+
+    model_weights_size = dtype_size * model_param_count_b * 1_000_000_000
+
+    min_memory_gb = math.ceil((min_kv_cache_size + model_weights_size) / 1_000_000_000 / 0.9)
+
+    print(
```
yixu34, line 2240: Log?

yixu34 (commented):
1. Cleanup prints
2. Unrelated changes?

model-engine/model_engine_server/infra/gateways/abs_llm_artifact_gateway.py
yunfeng-scale: need to add this function since i updated `LLMArtifactGateway`. not quite sure how to test though

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: lol yes

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2200,14 +2173,107 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
...
+
+    model_weights_size = dtype_size * model_param_count_b * 1_000_000_000
+
+    min_memory_gb = math.ceil((min_kv_cache_size + model_weights_size) / 1_000_000_000 / 0.9)
+
+    logger.info(
+        f"Memory calculation result: {min_memory_gb=} for {model_name}, min_kv_cache_size: {min_kv_cache_size}, model_weights_size: {model_weights_size}"
+    )
+
+    if min_memory_gb <= 24:
```
seanshi-scale, line 2246: probably not really an issue, but how well do these map to Azure instance types?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2200,14 +2173,107 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
...
+
+@lru_cache()
+def _infer_hardware(
+    llm_artifact_gateway: LLMArtifactGateway,
+    model_name: str,
+    checkpoint_path: str,
+) -> CreateDockerImageBatchJobResourceRequests:
+    config = llm_artifact_gateway.get_model_config(checkpoint_path)
+
+    dtype_size = 2
```
seanshi-scale, line 2215: guess we're gonna handle quantization later?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: with my limited experiences on Azure, they do provide the same set of GPUs @squeakymouse wdyt?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: quantization can already be handled here. but I chose not to update `dtype_size` since in my experiences, they usually make things slower not faster (at least for `bitsandbytes` and `AWQ`), so in order to achieve the same speed we still need the same amount of GPUs

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2200,14 +2173,107 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
...
+
+    min_kv_cache_size = (
```
seanshi-scale, line 2217: IIUC we're implicitly setting this to "batch size = 1 and filling up the context window", this feels reasonable, but would it make sense to add a bit of room for a larger batch size? (I guess for something with a shorter context window e.g. llama 2 it makes more sense to add some room, maybe less so for mixtral)

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -318,9 +295,9 @@ def validate_num_shards(
             raise ObjectHasInvalidValueException(
                 f"Num shard {num_shards} must be the same as number of GPUs {gpus} for DeepSpeed."
             )
-    if num_shards > gpus:
+    if num_shards != gpus:
```
seanshi-scale, line 298: should we just deprecate the `num_shards` field at this point?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2200,14 +2173,107 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
...
+    ):
+        if not (
+            request.gpus is None
+            and request.gpu_type is None
+            and request.cpus is None
+            and request.memory is None
+            and request.storage is None
+        ):
+            raise ObjectHasInvalidValueException(
+                "All hardware spec fields (gpus, gpu_type, cpus, memory, storage) must be provided if any hardware spec field is missing."
```
seanshi-scale, line 2194: nit: "... if any hardware spec field is provided"?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: for all the existing models i tested, batch size = 1 is a good enough default to reach to reasonable amount of GPUs. model builders would have thought about the tradeoffs between model sizes and GPU sizes

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: yes we should but would prefer not in this PR

seanshi-scale (approved):
had a few questions/nits, but lgtm!

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: created https://linear.app/scale-epd/issue/MLI-2315/deprecate-num-shards-field

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: err, i think my previous comment is incorrect https://github.com/scaleapi/llm-engine/pull/515#discussion_r1597247261

yunfeng-scale: https://github.com/scaleapi/llm-engine/pull/515#discussion_r1597247261
i guess it works both ways. since here i'm only allowing two states: either all fields are provided, or none of them are provided

PR #517, opened 2024-05-14
Increase max gpu utilization for 70b models

dmchoiboi:
# Pull Request Summary

Up max gpu memory utilization to 0.95 for 70b models in attempt to address OOM issues

https://linear.app/scale-epd/issue/MLI-2309/use-095-gpu-memory-utilization-for-70b-models

## Test Plan and Usage Guide

Published test docker image for batch_inference. Tested with API request using local gateway: job `ft-cp21h54gfe6g02mlqikg`

### DIFF
--- a/model-engine/model_engine_server/common/dtos/llms.py
+++ b/model-engine/model_engine_server/common/dtos/llms.py
@@ -528,15 +528,35 @@ class CreateBatchCompletionsRequest(BaseModel):
     """
     Maximum runtime of the batch inference in seconds. Default to one day.
     """
-    max_gpu_memory_utilization: Optional[float] = Field(default=0.9, le=1.0)
-    """
-    Maximum GPU memory utilization for the batch inference. Default to 90%.
-    """
     tool_config: Optional[ToolConfig] = None
     """
     Configuration for tool use.
     NOTE: this config is highly experimental and signature will change significantly in future iterations.
     """
+
+
+class CreateBatchCompletionsEngineRequest(CreateBatchCompletionsRequest):
+    """
+    Internal model for representing request to the llm engine. This contains additional fields that we want
+    hidden from the DTO exposed to the client.
+    """
+
+    max_gpu_memory_utilization: Optional[float] = Field(default=0.9, le=1.0)
+    """
+    Maximum GPU memory utilization for the batch inference. Default to 90%.
+    """
+
+    @staticmethod
+    def from_api(request: CreateBatchCompletionsRequest) -> "CreateBatchCompletionsEngineRequest":
+        return CreateBatchCompletionsEngineRequest(
+            input_data_path=request.input_data_path,
+            output_data_path=request.output_data_path,
+            content=request.content,
+            model_config=request.model_config,
+            data_parallelism=request.data_parallelism,
+            max_runtime_sec=request.max_runtime_sec,
+            tool_config=request.tool_config,
+        )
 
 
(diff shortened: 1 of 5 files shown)


model-engine/model_engine_server/common/dtos/llms.py
```diff
@@ -528,6 +528,10 @@ class CreateBatchCompletionsRequest(BaseModel):
     """
     Maximum runtime of the batch inference in seconds. Default to one day.
     """
+    max_gpu_memory_utilization: Optional[float] = Field(default=0.9, le=1.0)
```
yunfeng-scale, line 531: consider removing this. we have been trying to make batch completion params as simple as possible so users don't need to specify things like gpu/cpu count

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2323,6 +2344,12 @@ async def execute(
                 "Only code_evaluator tool is supported for batch completions."
             )
 
+        additional_engine_args = infer_addition_engine_args_from_model_name(
+            request.model_config.model
+        )
+        if additional_engine_args.gpu_memory_utilization is not None:
+            request.max_gpu_memory_utilization = additional_engine_args.gpu_memory_utilization
```
yunfeng-scale, line 2351: instead of putting this in request, can you update the bundle config?

dmchoiboi: It looks like the bundle isn't actually being used outside the context of this request, so it seems simpler to have a separate DTO for the Gateway vs the engine.

model-engine/model_engine_server/common/dtos/llms.py
dmchoiboi: Created a separate DTO for engine request

yunfeng-scale:
lgtm

yunfeng-scale (approved):

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2198,6 +2199,27 @@ async def execute(self, user: User, request: ModelDownloadRequest) -> ModelDownl
         return ModelDownloadResponse(urls=urls)
 
 
+@dataclass
+class VLLMEngineArgs:
```
yixu34, line 2203: Hm I know this is by no means the main offender, but implementation specifics like vLLM aren't supposed to go into the use case layer. Granted, that'd require another layer, which I suspect @yunfeng-scale would find perfunctory 😁

dmchoiboi: I guess I could just call it LLMEngineArgs. It seems right now we only support batch inference w/ vLLM, so we could try to do a proper abstraction when we decide we need to support it for a different engine?

yixu34: Yeah I think this is ok for now.

yunfeng-scale: oh 😅 you had a good point, the current code structure does not completely fit into clean architecture. in that sense we might want to move all these framework-specific code to another layer

PR #530, opened 2024-05-31
Remove Gunicorn and use Uvicorn only for gateway

yunfeng-scale:
# Pull Request Summary

## Test Plan and Usage Guide


model-engine/model_engine_server/entrypoints/start_fastapi_server.py
```diff
@@ -3,28 +3,42 @@
...
+        # with Windows or PyPy.
+        "--loop",
+        "uvloop",
+        "--http",
+        "httptools",
+        "--limit-concurrency",
+        f"{CONCURRENCY_LIMIT}",
         "--workers",
-        f"{num_workers}",
+        "1",  # Let the Kubernetes deployment handle the number of pods
```
edgan8, line 41: Will this reduce the amount of traffic we can receive per-pod? Why not keep it at 4?

edgan8 (commented):
This is a big change, can you include a summary of the test plan? I'm worried about any performance impact

model-engine/model_engine_server/entrypoints/start_fastapi_server.py
yunfeng-scale: this is to remove load balancing within pod

seanshi-scale: are we increasing the number of pods by 4 to compensate?

yunfeng-scale: that's the initial plan

yunfeng-scale: tbh i think llm engine is overprovisioned

yixu34:
> This is a big change, can you include a summary of the test plan? I'm worried about any performance impact

In addition to:
> will run simple load testing for get requests with and without gunicorn

Let's also monitor post-rollout 👀

PR #534, opened 2024-06-04
Modify v1 completions_stream logic to raise most exceptions before async streaming inference response

anant-marur:
# Pull Request Summary

Modifies `CompletionStreamV1UseCase.execute()` to create an asynchronous response chunk generator and return a reference to it, rather than acting as a generator itself. This enables `create_completion_stream_task()` to call `CompletionStreamV1UseCase.execute()` with `await`, enabling exception handling to occur before response streaming begins. 

This enables all exceptions that are raised prior to response streaming to be propagated back to the user as HTTPExceptions with the proper status codes, increasing clarity and usability of the system. Currently, all exceptions have to be wrapped as error messages in regular response chunks, returning a `200` status code regardless of exceptions.

There are still several exceptions that can be raised once the response has begun - these exceptions (`InvalidRequestException`, `UpstreamServerError`) will still be wrapped in `CompletionsStreamV1Response` objects.

## Test Plan and Usage Guide

### **Unit Testing**
```
$ just unit-test tests/unit/domain/test_llm_use_cases.py

...

================= 53 passed, 3 warnings in 7.35s =================
```

### **Local Server Testing**
Initialize server with `LOCAL=true just up`

1. Nonexistent endpoint names now raise a `HTTP 404`

```
$ curl -X POST -i \                                                                                                                       
     --url 'http://localhost:5001/v1/llm/completions-stream?model_endpoint_name=fake-llama-2-7b-anantmarur-test' \                                  
     --header 'accept: application/json' \                                                                                                          
     --header 'content-type: application/json' \                                                                                                    
     --header 'x-api-key: 0eefc220-128e-4031-9602-1a6b1132233d' \                                                                                   
     --data '                                                                                                                                       
{                                                                                                                                                   
  "temperature": 0.2, "max_new_tokens": 4096,                                                                                                       
  "stream": true,                                                                                                                                   
  "prompt": "[INST] Your task is to write 2 tests to check the correctness of a function that solves a programming problem. The tests must be betwee
n [TESTS] and [/TESTS] tags. You must write the comment \"#Test case n:\" on a separate line directly above each assert statement, where n represent
s the test case number, starting from 1 and increasing by one for each subsequent test case.   Problem: Write a Python function to get the unique el
ements of a list. [/INST]"                                                                                                                          
}                                                                                                                                                   
' -u 661595e1df6a6a51c8562e92                                                                                                                       
Enter host password for user '661595e1df6a6a51c8562e92':                                                                                            
HTTP/1.1 404 Not Found                                                                                                                              
date: Wed, 05 Jun 2024 00:39:14 GMT                                                                                                                 
server: uvicorn                                                                                                                                     
content-length: 70                                                                                                                                  
content-type: application/json                                                                                                                      
                                                                                                                                                    
{"detail":"Model endpoint fake-llama-2-7b-anantmarur-test not found."}
```

2. Issues with inference API call raise a wrapped `HTTP 500`. (Specifies an endpoint with no workers to spin up).

```
$ curl -X POST -i \                
                                                                                                                                                    
     --url 'http://localhost:5001/v1/llm/completions-stream?model_endpoint_name=codellama-7b' \                                                     
                                                                                                                                                    
     --header 'accept: application/json' \                                                                                                          
     --header 'content-type: application/json' \                                                                                                    
     --header 'x-api-key: 0eefc220-128e-4031-9602-1a6b1132233d' \                                                                                   
     --data '                                                                                                                                       
{                                                                                                                                                   
  "temperature": 0.2, "max_new_tokens": 128,                                                                                                        
  "stream": true,                                                         
  "prompt": "[INST] Your task is to write 2 tests to check the correctness of a function that solves a programming problem. The tests must be betwee
n [TESTS] and [/TESTS] tags. You must write the comment \"#Test case n:\" on a separate line directly above each assert statement, where n represent
s the test case number, starting from 1 and increasing by one for each subsequent test case.   Problem: Write a Python function to get the unique el
ements of a list. [/INST]"
}                                                                         
' -u 661595e1df6a6a51c8562e92                                             
Enter host password for user '661595e1df6a6a51c8562e92':                  
HTTP/1.1 200 OK                                                           
date: Wed, 05 Jun 2024 00:40:10 GMT                                       
server: uvicorn                                                           
cache-control: no-cache                                                   
connection: keep-alive                                                    
x-accel-buffering: no                                                     
content-type: text/event-stream; charset=utf-8                            
transfer-encoding: chunked

...

data: {"request_id": "5eafbb29-5f8a-4edc-bf3b-0f471693a0b6", "output": null, "error": {"status_code": 500, "content": {"error": "Upstream service er
ror for request_id 5eafbb29-5f8a-4edc-bf3b-0f471693a0b6", "timestamp": "2024-06-04 19:13:56 PDT"}}}
```

3. Streaming still works as intended for non-problematic requests:
```
$ curl -X POST -i \                                                                                                                                                                                          
     --url 'http://localhost:5001/v1/llm/completions-stream?model_endpoint_name=llama-2-7b-anantmarur-test' \                                                                                                          
     --header 'accept: application/json' \                                                                                                                                                                             
     --header 'content-type: application/json' \                                                                                                                                                                       
     --header 'x-api-key: 0eefc220-128e-4031-9602-1a6b1132233d' \                                                                                                                                                      
     --data '                                                                                                                                                                                                          
{                                                                                                                                                                                                                      
  "temperature": 0.2, "max_new_tokens": 64,                                                                                                                                                                            
  "stream": true,                                                                                                                                                                                                      
  "prompt": "The biggest health benefits of avocados are "                                                                                          }                                                                  
' -u 661595e1df6a6a51c8562e92                                                                                                                                                                                          
Enter host password for user '661595e1df6a6a51c8562e92':                                                                                                                                                               
HTTP/1.1 200 OK                                                                                                                                                                                                        
date: Wed, 05 Jun 2024 05:33:19 GMT                                                                                                                                                                                    
server: uvicorn                                                                                                                                                                                                        
cache-control: no-cache                                                                                                                                                                                                
connection: keep-alive                                                                                                                                                                                                 
x-accel-buffering: no                                                                                                                                                                                                  
content-type: text/event-stream; charset=utf-8                                                                                                                                                                         
transfer-encoding: chunked                                                                                                                                                                                             
                                                                                                                                                                                                                       
data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": "1", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 1, "token": null}, "error": null}                          
                                                                                                                                                                                                                       
data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": ".", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 2, "token": null}, "error": null}                          
                                                                                                                                                                                                                       
data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": " Av", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 3, "token": null}, "error": null}                        
                                                                                                                                                                                                                       
data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": "oc", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 4, "token": null}, "error": null}                         
                                                                                                                                                                                                                       
data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": "ados", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 5, "token": null}, "error": null}                       
                                                                                                                                                                                                                       
data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": " are", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 6, "token": null}, "error": null}                       
                                                                                                                                                                                                                       
data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": " high", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 7, "token": null}, "error": null}                      

data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": " in", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 8, "token": null}, "error": null}

...


data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": " ch", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 61, "token": null}, "error": null}                       

data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": "ol", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 62, "token": null}, "error": null}                        

data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": "ester", "finished": false, "num_prompt_tokens": null, "num_completion_tokens": 63, "token": null}, "error": null}                     

data: {"request_id": "4cbcc973-c191-49db-bb8e-eed7d69be08f", "output": {"text": "ol", "finished": true, "num_prompt_tokens": 11, "num_completion_tokens": 64, "token": null}, "error": null}                           
```


model-engine/model_engine_server/api/llms_v1.py
```diff
@@ -389,7 +389,28 @@ async def create_completion_stream_task(
         llm_model_endpoint_service=external_interfaces.llm_model_endpoint_service,
         tokenizer_repository=external_interfaces.tokenizer_repository,
     )
-    response = use_case.execute(user=auth, model_endpoint_name=model_endpoint_name, request=request)
+    # Call execute() synchronously, execute will asynchronously call response streaming logic and return the response object
```
yunfeng-scale, line 392: does this actually work? can you test it out with some endpoint?

anant-marur: Revised by adding `await`, confirmed this works in the test plan. @edgan8 had some concerns about this approach, given it changes the semantics of `execute()` to now be called with `await`, although I think it seems okay given it lines up with the use of `await` in all other instances of `execute()` in this file. WDYT?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1827,18 +1827,27 @@ async def execute(
     ) -> AsyncIterable[CompletionStreamV1Response]:
         """
         Runs the use case to create a stream inference task.
+        NOTE: Must be called with await(), since the function is not a generator itself, but rather creates one and
```
seanshi-scale, line 1830: qq: I'm slightly surprised that the type that's returned doesn't change, since we go from not awaiting the result of this to awaiting the result of this. (unless the type that's returned does change)

edgan8: I think the new type is something like `Awaitable[AsyncIterable[CompletionStreamV1Response]]` now, which is a pretty complex type. I'm also worried that the new inline function will introduce some subtle bugs. 

I think it would be better to just move the exception throwing logic into a separate `async def validate()` method, and minimize the amount of code changes to the execute method, but open to suggestions from people more familiar with this code.

anant-marur: The return type actually doesn't change. 

TL;DR: `execute()` has to be awaited after the change because it `return`s a value, instead of `yield`ing values, changing its semantics from an asynchronous generator to a regular asynchronous function (outsourcing the asynchronous generator work to `_response_chunk_generator` in the new version.) In both cases, it returns an `AsyncIterable` to be used in an `async for` loop. 

Details:

From the Python docs, an [asynchronous generator](https://docs.python.org/3.8/glossary.html#term-asynchronous-generator) is:

> A function which returns an [asynchronous generator iterator](https://docs.python.org/3.8/glossary.html#term-asynchronous-generator-iterator). **It looks like a coroutine function defined with [async def](https://docs.python.org/3.8/reference/compound_stmts.html#async-def) except that it contains [yield](https://docs.python.org/3.8/reference/simple_stmts.html#yield) expressions** for producing a series of values usable in an [async for](https://docs.python.org/3.8/reference/compound_stmts.html#async-for) loop.

Note that asynchronous generators are never called with `await`, since their execution is contingent on being processed in the context of an `async for`.

And, an [asynchronous generator iterator](https://docs.python.org/3.8/glossary.html#term-asynchronous-generator-iterator) is:

> **An object created by a [asynchronous generator](https://docs.python.org/3.8/glossary.html#term-asynchronous-generator) function.**
This is an [asynchronous iterator](https://docs.python.org/3.8/glossary.html#term-asynchronous-iterator) which when called using the [\_\_anext\_\_()](https://docs.python.org/3.8/reference/datamodel.html#object.__anext__) method returns an awaitable object which will execute the body of the asynchronous generator function until the next [yield](https://docs.python.org/3.8/reference/simple_stmts.html#yield) expression.


So, before the fix:

- `CompletionStreamV1UseCase.execute()` **is an asynchronous generator** since it is defined with `async def` and contains `yield` statements. This means it returns an asynchronous generator iterator object by definition (`AsyncIterable[CompletionStreamV1Response]` in this case) when called without `await`.

And, after the fix:
- `_response_chunk_generator()` **is an asynchronous generator** since it is defined with `async def` and contains `yield` statements. So, similar to above, it returns `AsyncIterable[CompletionStreamV1Response]` when called without `await.
- `CompletionStreamV1UseCase.execute()` **is NOT an asynchronous generator**, since it no longer contains any `yield` statements. Rather, it concludes with `return _response_chunk_generator()`, calling `_response_chunk_generator()` without `await` and directly returning that object, which is `AsyncIterable[CompletionStreamV1Response]`. Since `execute()` **returns** now, rather than **yielding**, it is a regular async function rather than a generator, and has to be `await`ed to obtain its return value.

anant-marur: I'll wave the white flag on this approach if that doesn't clarify things - I already have @edgan8's suggestion mostly complete anyway. However, I think there are few advantages of this approach over the `validate()` approach:

- If we don't want to keep track of which endpoint names have been validated within the use case class state, `validate()` must return the fully processed model_content to `create_completion_stream_task`, which will pass it to `execute()`. `execute()` then has to assume `validate()` has been called and perform no additional validation on the model content.
- Otherwise, the use case must store the fully processed model content, (or a set of validated endpoint names, but this would mean execute() has to fetch the endpoints again). and execute has to read this state before initializing streaming. Otherwise, the risk exists where execute() can be called on an unvalidated endpoint.
- separating out a `validate()` function also creates a gap in execution between completing validation of the request, and the beginning of inference streaming. In the old implementation, as well as this initial approach, there is no gap in control flow between the validation of the request and beginning inference.

Some of these might be actual concerns, others might just be me nitpicking - but these are the things that bothered me a bit while implementing the alternative. Would love to hear what all of your thoughts are!

seanshi-scale: IIUC the typical pattern inside of LLM Engine's code is that `use_case.execute()` is called from a different layer, and it can call various helpers e.g. a hypothetical `_validate()`. Would that work? (not sure if that's what @edgan8 is going for)

model-engine/tests/unit/api/test_llms.py
```diff
@@ -156,15 +156,13 @@ def test_completion_sync_endpoint_not_found_returns_404(
     assert response_1.status_code == 404
 
 
-# When enabling this test, other tests fail with "RunTumeError got Future <Future pending> attached to a different loop"
-# https://github.com/encode/starlette/issues/1315#issuecomment-980784457
-@pytest.mark.skip(reason="Need to figure out FastAPI test client asyncio funkiness")
-def test_completion_stream_success(
+@pytest.mark.asyncio
```
edgan8, line 159: Is it safe to enable this test? What happened to the previous issue?

model-engine/tests/unit/api/test_llms.py
```diff
@@ -220,17 +219,13 @@ def test_completion_stream_endpoint_not_found_returns_404(
...
         json=completion_stream_request,
-        stream=True,
-    )
-
-    assert response_1.status_code == 200
-
-    for message in response_1:
-        assert "404" in message.decode("utf-8")
+    ) as r:
+        assert r.status_code == 404
```
edgan8, line 228: Great!

model-engine/tests/unit/api/test_llms.py
anant-marur: after modifying the body of the test to leverage `async` ContextManagers (ensuring that any asynchronous generators properly call `aclose` prior to garbage collection, which I suspect may have been the issue), and leveraging the new `get_async_test_client_wrapper` fixture I defined in `conftest.py`, I was able to run `just unit-test` on this entire file with no issues.

anant-marur: Ah, nvm - just noticed its affecting tests defined in other files, I misunderstood the initial comment.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2011,61 +2019,129 @@ async def execute(
...
-                            result["result"]["response"][0]["token_probs"]["tokens"]
-                        )
-                        yield CompletionStreamV1Response(
-                            request_id=request_id,
-                            output=CompletionStreamOutput(
-                                text=result["result"]["response"][0]["text"],
-                                finished=True,
-                                num_prompt_tokens=num_prompt_tokens,
-                                num_completion_tokens=completion_token_count,
+        async def _response_chunk_generator() -> AsyncIterable[CompletionStreamV1Response]:
```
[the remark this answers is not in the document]
seanshi-scale: I add `?w=1` to ignore the whitespace changes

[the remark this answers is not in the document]
edgan8: ohh amazing, if that works then my comment is less necessary.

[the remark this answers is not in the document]
anant-marur: Implemented the change anyway to make execute() less messy

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2011,61 +2019,129 @@ async def execute(
...
+            # inform interpreter to reference num_prompt_tokens defined in outer scope during assignment
+            nonlocal num_prompt_tokens
```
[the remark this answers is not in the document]
anant-marur: done.

model-engine/tests/unit/api/test_llms.py
anant-marur: I fixed the weird interaction between these tests by defining an event_loop fixture in `tests/unit/api/conftest.py` that ensures each async test runs with its own event loop, and reconfiguring `test_create_streaming_task_success` as an async test. cc @yunfeng-scale, I think you were the last to take a look at this issue?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2011,15 +2022,55 @@ async def execute(
...
             topic=model_endpoint.record.destination, predict_request=inference_request
         )
 
         num_completion_tokens = 0
         async for res in predict_result:
-            result = res.result
-            if model_content.inference_framework == LLMInferenceFramework.DEEPSPEED:
-                if res.status == TaskStatus.SUCCESS and result is not None:
+            if not res.status == TaskStatus.SUCCESS or res.result is None:
+                # Raise an UpstreamServiceError if the task has failed
```
edgan8, line 2057: I couldn't find any logic in the existing code that would throw this error on task failure, did you introduce this behavior? It does seem like a usability improvement.

edgan8 (approved):
This change looks very clean now, the new error handling is lovely! Glad you were able to figure out how to revive the old tests as well. LGTM, though will defer to Sean and Yunfeng to double check the core logic.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
anant-marur: The streaming model inference gateway will propagate a `FAILURE` status back from [here](https://github.com/scaleapi/llm-engine/blob/69163b23a3d63e0ae34a7fd55cd1f9de439c8cca/model-engine/model_engine_server/infra/gateways/live_streaming_model_endpoint_inference_gateway.py#L223).

edgan8: Sure, so this is another instance of a new exception we are throwing that was previously encoded in the streaming response right?

Let's document all of these so we can make an announcement that the endpoint is now throwing more helpful exceptions in X following cases.

seanshi-scale (approved):
looks good! also thanks for fixing the async unit tests

model-engine/tests/unit/api/conftest.py
```diff
@@ -90,6 +93,14 @@ def fake_auth():
         app.dependency_overrides[verify_authentication] = {}
 
 
+@pytest_asyncio.fixture(scope="session", autouse=True)
```
yunfeng-scale, line 96: thanks for fixing this!

model-engine/tests/unit/api/test_llms.py
yunfeng-scale: lgtm!

yunfeng-scale (approved):
thanks for fixing our stream serving!

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: thanks for the explanations! i'm still a bit confused why the signature isn't changed to `Awaitable[AsyncIterable[CompletionStreamV1Response]]`? also curious how does Python decide if a function is async generator? does it execute the function up until a `yield` is met and suddenly the value needs to be received by `async for`?

edgan8: This is my understanding: 

async `functions` are annotated with return type X, but actually return Awaitable[X]. Async generators are annotated with AsyncIterable[X] and actually return AsyncIterable[X]. This leads to confusion around the type signature because it's hard to tell if something is a function or a generator until you find the `yield` or `return`.

Python decides if a function is an async generator if it has a yield instead of a return. I am also curious if that is done statically ahead of time or line-by-line.

PR #543, opened 2024-06-18
Infer hardware specs from config

yunfeng-scale:
# Pull Request Summary

when inferring hardware specs, use a configmap to store the config instead of hardcode

## Test Plan and Usage Guide

running locally to create endpoint in cluster


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2272,50 +2290,29 @@ def _infer_hardware(
...
+        for recs in by_gpu_memory_gb:
+            if min_memory_gb <= recs["gpu_memory_le"]:
+                cpus = recs["cpus"]
+                gpus = recs["gpus"]
+                memory = recs["memory"]
+                storage = recs["storage"]
+                gpu_type = recs["gpu_type"]
+                found = True
+                break
+        if not found:
```
seanshi-scale, line 2314: nit: do we want to use `for/else` here? to get rid of the `found` flag

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2272,50 +2290,29 @@ def _infer_hardware(
...
-        memory = "800Gi"
-        storage = "460Gi"
-        gpu_type = GpuType.NVIDIA_HOPPER_H100
-    elif "llama-3-8b-instruct-262k" in model_name:
-        cpus = "20"
-        gpus = 2
-        memory = "40Gi"
-        storage = "40Gi"
-        gpu_type = GpuType.NVIDIA_HOPPER_H100
+    config_map = await _get_recommended_hardware_config_map()
```
seanshi-scale, line 2293: do we want to have a fallback if this config_map isn't set? it looks like we need to have the helm value `recommendedHardware` set, otherwise I think inferring hardware breaks, and batch completions won't work

seanshi-scale (commented):
had a question about how batch inference would work + a nit, looks good otherwise

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: i can't foresee how we can come up with anything if it's not set. do you mean keep some fallback logics in code?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
yunfeng-scale: good point, thanks!

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
seanshi-scale: yeah that's one way to go about it, i.e. right now I don't think there's an option to pass hardware choices to batch inference, so we'd have to either let the user provide that info, or just say "you have to set that recommendHardware thing if you want to use batch inference", which might be ok actually

yunfeng-scale: i think putting example values in values_sample.yaml should be good enough?

seanshi-scale: yeah that should be fine

seanshi-scale (approved):

PR #544, opened 2024-06-18
Add special token param to completions + batch completions apis

seanshi-scale:
# Pull Request Summary

Adds an option to show the special tokens in completions/stream completions/batch completions output

## Test Plan and Usage Guide

Completions/Stream completions
- Deploy a model (or grab one from somewhere) that has the special tokens
- Make request to model to see if special tokens are respected (with skip = False). Also try skip = True to sanity check

Made request with body ```{
"prompt": "Give me a list of all numbers. Sure! 1, 2",
"max_new_tokens": 10,
"temperature": 0.0,
"skip_special_tokens": false,
"guided_regex": "4"
}```, this gives output "4</s>", making same request with skip=true, output is just "4". Works for sync and stream.

Batch completions
- Make batch completions call using model with special tokens
- See whether output has the special tokens

Made batch completions calls with both skip special token=true and false, true had empty spaces where tokens should have been, false had the special tokens.

### DIFF
--- a/model-engine/model_engine_server/inference/batch_inference/build_and_upload_image.sh
+++ b/model-engine/model_engine_server/inference/batch_inference/build_and_upload_image.sh
@@ -17,5 +17,5 @@ fi
 IMAGE_TAG=$2
 ACCOUNT=$1
 aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin $ACCOUNT.dkr.ecr.us-west-2.amazonaws.com
-DOCKER_BUILDKIT=1 docker build -t $ACCOUNT.dkr.ecr.us-west-2.amazonaws.com/llm-engine/batch-infer-vllm:$IMAGE_TAG -f Dockerfile_vllm ../../../../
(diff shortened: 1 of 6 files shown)


model-engine/model_engine_server/inference/batch_inference/build_and_upload_image.sh
```diff
@@ -18,4 +18,4 @@ IMAGE_TAG=$2
 ACCOUNT=$1
 aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin $ACCOUNT.dkr.ecr.us-west-2.amazonaws.com
 DOCKER_BUILDKIT=1 docker build -t $ACCOUNT.dkr.ecr.us-west-2.amazonaws.com/llm-engine/batch-infer-vllm:$IMAGE_TAG -f Dockerfile_vllm ../../../../
-docker push $ACCOUNT.dkr.ecr.us-west-2.amazonaws.com/llm-engine/batch-infer-vllm:$IMAGE_TAG
+docker push $ACCOUNT.dkr.ecr.us-west-2.amazonaws.com/llm-engine/batch-infer-vllm-test:$IMAGE_TAG
```
seanshi-scale, line 21: TODO revert this

yunfeng-scale (approved):

PR #545, opened 2024-06-20
Mitigation for AsyncEngineDeadError

dmchoiboi:
# Pull Request Summary

There is a known error where vLLM's AsyncEngine will fail silently on OOM, but the webserver will continue running. There was a previous attempt at fixing this, but it didn't fully capture the line the error was happening. 

Stack trace:
```
File "/usr/local/lib/python3.10/dist-packages/fastapi/routing.py", line 191, in run_endpoint_function
  return await dependant.call(**values)
File "/workspace/vllm_server.py", line 112, in generate
  async for request_output in results_generator:
File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 666, in generate
  raise e
File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 650, in generate
  stream = await self.add_request(
File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 537, in add_request
  self.start_background_loop()
File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 411, in start_background_loop
raise AsyncEngineDeadError(
  vllm.engine.async_llm_engine.AsyncEngineDeadError: Background loop has errored already.
```

To mitigate this, we want to run the engine healthcheck on every request.
To kill the server, we have to manually kill the process via os since raising sigterm is caught by some middleware.

## Test Plan and Usage Guide

Tested with sample uvicorn + fastapi server throwing deadengine errors

```
from typing import Any
import uvicorn
import signal
import os

from fastapi import BackgroundTasks, FastAPI, HTTPException, Request
from fastapi.responses import Response

app = FastAPI()

class AsyncEngineDeadError(Exception):
    pass

async def throw_error():
    raise AsyncEngineDeadError()

@app.get("/predict")
async def generate(request: Request) -> Response:
    try:
        await throw_error()
    except Exception as e:
        print(f"Engine not healthy: {e}")
        os.kill(os.getpid(), signal.SIGINT)

if __name__ == "__main__":
    uvicorn.run(app, port=8080, log_level="debug")

```

### DIFF
--- a/model-engine/model_engine_server/inference/vllm/vllm_server.py
+++ b/model-engine/model_engine_server/inference/vllm/vllm_server.py
@@ -1,6 +1,7 @@
 import argparse
 import code
 import json
+import os
 import signal
 import subprocess
 import traceback
@@ -44,7 +45,7 @@ async def generate(request: Request) -> Response:
         await engine.check_health()
     except Exception as e:
         print(f"The vllm engine is dead, exiting the pod: {e}")
-        exit(1)
+        os.kill(os.getpid(), signal.SIGINT)
 
     request_dict = await request.json()
     prompt = request_dict.pop("prompt")
--- a/model-engine/model_engine_server/inference/vllm/vllm_server.py
+++ b/model-engine/model_engine_server/inference/vllm/vllm_server.py
@@ -10,7 +10,7 @@ import uvicorn
 from fastapi import BackgroundTasks, FastAPI, HTTPException, Request
 from fastapi.responses import Response, StreamingResponse
 from vllm.engine.arg_utils import AsyncEngineArgs
-from vllm.engine.async_llm_engine import AsyncEngineDeadError, AsyncLLMEngine
+from vllm.engine.async_llm_engine import AsyncLLMEngine
 from vllm.entrypoints.openai.protocol import CompletionRequest as OpenAICompletionRequest
 from vllm.model_executor.guided_decoding import get_guided_decoding_logits_processor
 from vllm.outputs import CompletionOutput
@@ -39,6 +39,13 @@ async def generate(request: Request) -> Response:
     - stream: whether to stream the results or not.
     - other fields: the sampling parameters (See `SamplingParams` for details).
     """
+    # check health before accepting request and fail fast if engine isn't healthy
+    try:
+        await engine.check_health()
+    except Exception as e:
+        print(f"The vllm engine is dead, exiting the pod: {e}")
+        exit(1)
+
     request_dict = await request.json()
     prompt = request_dict.pop("prompt")
     stream = request_dict.pop("stream", False)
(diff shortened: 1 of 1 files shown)


model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -77,57 +77,57 @@ async def generate(request: Request) -> Response:
...
+        prompt = final_output.prompt
+        ret = {
+            "text": final_output.outputs[0].text,
+            "count_prompt_tokens": len(final_output.prompt_token_ids),
+            "count_output_tokens": len(final_output.outputs[0].token_ids),
+            "log_probs": format_logprobs(final_output),
+            "tokens": tokens,
+        }
+        return Response(content=json.dumps(ret))
+    except AsyncEngineDeadError as e:
```
yunfeng-scale, line 128: @anant-marur i'm not sure if this would capture the streaming case. any suggestions?

model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -77,57 +77,57 @@ async def generate(request: Request) -> Response:
...
     try:
         results_generator = engine.generate(prompt, sampling_params, request_id)
-    except AsyncEngineDeadError as e:
-        print(f"The vllm engine is dead, exiting the pod: {e}")
-        exit(1)
 
-    # Streaming case
-    async def stream_results() -> AsyncGenerator[str, None]:
+        # Streaming case
+        async def stream_results() -> AsyncGenerator[str, None]:
```
yunfeng-scale, line 82: now i think the type annotation should be AsyncIterator?

dmchoiboi: I think AsyncGenerator is still correct?

yunfeng-scale (approved):

model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -39,6 +39,9 @@ async def generate(request: Request) -> Response:
     - stream: whether to stream the results or not.
     - other fields: the sampling parameters (See `SamplingParams` for details).
     """
+    # check health before accepting request and fail fast if engine isn't healthy
+    await engine.check_health()
```
seanshi-scale, line 43: does this exit the pod if it fails? or does it just fail the request

dmchoiboi: yeah i'm still testing this out, but this will just fail the request. I tried exit 1 as well but i don't think uvicorn itself exits if the underlying fastapi server exits so trying to fgiure out a solution for this

seanshi-scale: mb, it might actually be ok to just exit the process? as long as whatever is holding the gpu memory gets killed I think it ends up being ok

PR #548, opened 2024-06-24
Fix integration tests for streaming case

dmchoiboi:
# Pull Request Summary

vLLM integration tests for the stream case has a few bugs that this addresses

## Test Plan and Usage Guide

### DIFF
--- a/integration_tests/rest_api_utils.py
+++ b/integration_tests/rest_api_utils.py
@@ -168,7 +168,7 @@ INFERENCE_PAYLOAD: Dict[str, Any] = {
 
 CREATE_LLM_MODEL_ENDPOINT_REQUEST: Dict[str, Any] = {
     "name": format_name("llama-2-7b-test"),
-    "model_name": "llama-2-7b",
+    "model_name": "llama-2-7b-chat",
     "source": "hugging_face",
     "inference_framework": "vllm",
     "inference_framework_image_tag": "latest",
@@ -802,7 +802,7 @@ async def create_llm_streaming_task(
         timeout=LONG_NETWORK_TIMEOUT_SEC,
     ) as response:
         assert response.status == 200, (await response.read()).decode()
-        return await response.json()
+        return (await response.read()).decode()
 
 
 async def create_sync_tasks(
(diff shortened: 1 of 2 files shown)


integration_tests/rest_api_utils.py
```diff
@@ -987,6 +987,34 @@ def ensure_llm_task_response_is_correct(
         assert re.search(response_text_regex, response["output"]["text"])
 
 
+def ensure_llm_task_stream_response_is_correct(
+    response: str,
+    required_output_fields: Optional[List[str]],
+    response_text_regex: Optional[str],
+):
+    print("raw response", response)
```
yunfeng-scale, line 995: remove this?

integration_tests/rest_api_utils.py
```diff
@@ -802,7 +802,7 @@ async def create_llm_streaming_task(
         timeout=LONG_NETWORK_TIMEOUT_SEC,
     ) as response:
         assert response.status == 200, (await response.read()).decode()
-        return await response.json()
+        return (await response.read()).decode()
```
yunfeng-scale, line 805: hmm, how did this work before?

yunfeng-scale (approved):

integration_tests/rest_api_utils.py
dmchoiboi: i'm pretty sure it didn't

integration_tests/rest_api_utils.py
dmchoiboi: we can; it's been helpful for debugging if the tests fail

integration_tests/rest_api_utils.py
```diff
@@ -987,6 +987,34 @@ def ensure_llm_task_response_is_correct(
...
+    #  We want to get a list of dictionaries parsing out the 'data:' field
+    parsed_response = [
+        json.loads(r.split("data: ")[1]) for r in response.split("\n") if "data:" in r.strip()
+    ]
+
+    print("parsed response: ", parsed_response)
+    # if required_output_fields is not None:
+    #     for field in required_output_fields:
+    #         for response in parsed_response:
+    #             assert field in response["output"]
```
yunfeng-scale, line 1007: should implement these?

dmchoiboi: i'm not really sure what this is supposed to be testing. the response for streaming is different from that for sync so the same checks don't really apply

PR #550, opened 2024-06-26
Update vllm batch job to work with vllm > 0.5.0

dmchoiboi:
# Pull Request Summary

* Duplicated DTOs for batch creation to decouple pydantic versions
  * Had to rename model_config => model_cfg since pydantic 2.0 reserves `model_config` as a keyword
* Updates to handle vllm api changes

## Test Plan and Usage Guide


model-engine/model_engine_server/common/dtos/llms.py
```diff
@@ -553,6 +553,14 @@ class CreateBatchCompletionsEngineRequest(CreateBatchCompletionsRequest):
     hidden from the DTO exposed to the client.
     """
 
+    model_cfg: CreateBatchCompletionsModelConfig
+    """
+    Model configuration for the batch inference. Hardware configurations are inferred.
+
+    We rename model_config from api to model_cfg in engine since engine uses pydantic v2 which
+    reserves model_config as a keyword.
```
yunfeng-scale, line 561: what 😓

yunfeng-scale (approved):

model-engine/model_engine_server/inference/batch_inference/dto.py
```diff
@@ -0,0 +1,165 @@
+# This is a copy of model_engine_server.common.dtos.llm
```
seanshi-scale, line 1: should we have a pointer from model_engine_server.common.dtos.llm to here as well? so it's easier to remember to keep the two in sync

model-engine/model_engine_server/common/dtos/llms.py
```diff
@@ -565,6 +573,7 @@ def from_api(request: CreateBatchCompletionsRequest) -> "CreateBatchCompletionsE
             output_data_path=request.output_data_path,
             content=request.content,
             model_config=request.model_config,
+            model_cfg=request.model_config,
```
seanshi-scale, line 576: is there a reason why we have model_config and model_cfg?

dmchoiboi: it's kinda jank, but it's for forward/backwards compatibility.

On the API side, we want to keep model_config, but on the engine side, we want model_cfg

PR #560, opened 2024-07-09
Update vllm server to be openai compatible

dmchoiboi:
# Pull Request Summary

Update vllm server to be OpenAI compatible

A lot of this was nabbed from https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/api_server.py

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/inference/vllm/vllm_server.py
+++ b/model-engine/model_engine_server/inference/vllm/vllm_server.py
@@ -1,33 +1,86 @@
 import argparse
+import asyncio
 import code
 import json
+import logging
 import os
 import signal
 import subprocess
 import traceback
+from logging import Logger
 from typing import AsyncGenerator, Dict, List, Optional
 
 import uvicorn
 from fastapi import BackgroundTasks, FastAPI, HTTPException, Request
-from fastapi.responses import Response, StreamingResponse
+from fastapi.responses import JSONResponse, Response, StreamingResponse
 from vllm.engine.arg_utils import AsyncEngineArgs
 from vllm.engine.async_llm_engine import AsyncEngineDeadError, AsyncLLMEngine
+from vllm.entrypoints.openai.cli_args import make_arg_parser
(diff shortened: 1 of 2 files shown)


model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -200,23 +253,63 @@ def extract_logprobs(logprobs: Dict[int, Logprob]) -> Dict[int, float]:
...
+
+    signal.signal(signal.SIGUSR1, debug)
 
     engine_args = AsyncEngineArgs.from_cli_args(args)
     engine = AsyncLLMEngine.from_engine_args(engine_args)
 
-    signal.signal(signal.SIGUSR1, debug)
+    event_loop: Optional[asyncio.AbstractEventLoop]
+    try:
+        event_loop = asyncio.get_running_loop()
```
yunfeng-scale, line 285: when would there be event loop?

yunfeng-scale (approved):

model-engine/model_engine_server/inference/vllm/vllm_server.py
dmchoiboi: I guess for our use case, there wouldn't be one

model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -1,33 +1,86 @@
...
+    return Response(status_code=200)
+
+
+@app.get("/v1/models")
+async def show_available_models():
+    models = await openai_serving_chat.show_available_models()
+    return JSONResponse(content=models.model_dump())
+
+
+@app.post("/v1/chat/completions")
```
yunfeng-scale, line 63: how would you use these APIs though? do you have follow up PRs?

dmchoiboi: yes, will include in follow-up prs; just trying to keep these PRs small

PR #563, opened 2024-07-10
fix: Use env AWS_REGION in sqs_client or default to us-west-2

nicolastomeo:
# Pull Request Summary

Some customers deploy in different AWS regions

## Test Plan and Usage Guide

Pod has AWS_REGION so it should work or default to previous behaviour

### DIFF
--- a/model-engine/model_engine_server/core/celery/celery_autoscaler.py
+++ b/model-engine/model_engine_server/core/celery/celery_autoscaler.py
@@ -54,6 +54,7 @@ logger = make_logger(logger_name())
 
 autoscaler_broker = os.environ.get("BROKER_NAME", SQS_BROKER)
 aws_profile = os.environ.get("AWS_PROFILE")
+aws_region = os.environ.get("AWS_REGION", "us-west-2")
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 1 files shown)


model-engine/model_engine_server/core/celery/celery_autoscaler.py
```diff
@@ -54,6 +54,7 @@ def excluded_namespaces():
 
 autoscaler_broker = os.environ.get("BROKER_NAME", SQS_BROKER)
 aws_profile = os.environ.get("AWS_PROFILE")
+aws_region = os.environ.get("AWS_REGION", "us-west-2")
```
seanshi-scale, line 57: we do have `infra_config().default_region`, this gets (pretty much) passed through either through helm installs or through configs baked into the image (probably doesn't apply for your use case). I'm not sure if we have `AWS_REGION` as an available env var. cc @squeakymouse

nicolastomeo: I think it's currently injected by the IRSA setup

nicolastomeo: So this basically https://github.com/aws/amazon-eks-pod-identity-webhook injects the AWS_REGION in EKS cluster once you setup the service account with a role

seanshi-scale (approved):
lgtm!

PR #566, opened 2024-07-15
Parse wrapped sync endpoint error

yunfeng-scale:
# Pull Request Summary

sometimes requests are wrapped and so are error details. do one extra step to get the data out

## Test Plan and Usage Guide

with the help of @Tma2333-Scale, tested for exception within the API (verified that we can get traceback),

and for exception from Pydantic before reaching to the API:
image

### DIFF
--- a/model-engine/model_engine_server/infra/gateways/live_sync_model_endpoint_inference_gateway.py
+++ b/model-engine/model_engine_server/infra/gateways/live_sync_model_endpoint_inference_gateway.py
@@ -187,11 +187,14 @@ class LiveSyncModelEndpointInferenceGateway(SyncModelEndpointInferenceGateway):
             logger.error(f"Service error on sync task: {exc.content!r}")
             try:
                 error_json = orjson.loads(exc.content.decode("utf-8"))
-                result_traceback = (
-                    error_json.get("detail", {}).get("traceback")
-                    if isinstance(error_json, dict)
-                    else None
-                )
+                if not isinstance(error_json, dict):
+                    result_traceback = None
+                else:
+                    if "result" in error_json:
+                        error_json = error_json["result"]
+                    result_traceback = error_json.get("detail", {}).get(
+                        "traceback", "Failed to parse traceback"
+                    )
                 return SyncEndpointPredictV1Response(
                     status=TaskStatus.FAILURE,
                     traceback=result_traceback,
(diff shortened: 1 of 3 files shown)


model-engine/model_engine_server/infra/gateways/live_sync_model_endpoint_inference_gateway.py
```diff
@@ -186,12 +186,22 @@ async def predict(
...
-                    if isinstance(error_json, dict)
-                    else None
-                )
+                if not isinstance(error_json, dict):
+                    result_traceback = None
+                else:
+                    if "result" in error_json:
+                        error_json = error_json["result"]
+                        if not isinstance(error_json, dict):
+                            error_json = orjson.loads(error_json)
```
yixu34, line 197: Can this fail?

model-engine/model_engine_server/infra/gateways/live_sync_model_endpoint_inference_gateway.py
```diff
@@ -186,12 +186,22 @@ async def predict(
...
             logger.error(f"Service error on sync task: {exc.content!r}")
             try:
+                # Try to parse traceback from the response, fallback to just return all the content if failed
                 error_json = orjson.loads(exc.content.decode("utf-8"))
-                result_traceback = (
-                    error_json.get("detail", {}).get("traceback")
-                    if isinstance(error_json, dict)
-                    else None
-                )
+                if not isinstance(error_json, dict):
```
yixu34, line 191: Wonder if we can reduce nesting

model-engine/model_engine_server/infra/gateways/live_sync_model_endpoint_inference_gateway.py
yunfeng-scale: if this fails it'll go to L209 and we'll return everything unpacked

model-engine/model_engine_server/infra/gateways/live_sync_model_endpoint_inference_gateway.py
yunfeng-scale: there are 3 cases i'm checking:
detail.traceback
result.detail.traceback
result.detail[]

i could do simplification i guess

yixu34 (approved):

PR #574, opened 2024-07-23
Multinode serving

seanshi-scale:
# Pull Request Summary

Multinode serving

API/schema changes:
Bundles/bundle v2:
~Fields in metadata to allow for LWS, specifically we need a different command + some sets of env vars. No code really needed here iiuc since we're just passing through the existing metadata field.~
Use the new fields in the entity for doing multinode

Endpoints:
nodes_per_worker param that controls whether to use a LWS or not
Create LWS endpoint, delete LWS endpoint is allowed
Modifying a LWS endpoint to a non-LWS endpoint or modifying a non-LWS endpoint into a LWS endpoint is not allowed (similar to how we don't allow switching from sync and async) (also changing nodes_per_worker will not be allowed). Enforced by this not being capable through the API.
Get LWS endpoint is allowed
Only available LWS endpoint type will be gpu + streaming, cpu-only LWS doesn't really make sense imo, async/sync could make sense but we don't need it for LLM serving at this point
k8s_erd get_resources + get_all_resources are made LWS compatible
LWS endpoints create a service (separate from the service that the LWS auto-creates) + (if istio is enabled) a ServiceEntry. If istio is enabled, gateway will hackily manually resolve the service's IP and make a request there.

LLM endpoints:
create a multinode bundle if situation calls for it, use endpoint's api to include this new bundle
(TODO reflect in production) build new vllm image to use multinode
Endpoint will do tensorparallel=(# gpus in a single node), pipelineparallel=(# nodes)

Python client: 
Add way to set nodes_per_worker on create + get nodes_per_worker on get/list

## Test Plan and Usage Guide

unit tests
~TODO maybe test for updating existing LWS, _delete_lws doesn't find the lws to delete, _get_all has a LWS returned~
integration tests/e2e tests require LWS, so can't really do this
deploy and manually test some things
ie 
- create llm multinode works, (yes)
- update llm multinode replicas works, (yes)
- get endpoint/llm works, (yes)
- get all endpoint/llm works, (yes)
- delete llm multinode works, (yes)
- request to multinode llm endpoint works (yes)
  - ~currently `curl launch-endpoint-id-end-crtjkptlgk4g03v40o1g.scale-deploy.svc.cluster.local:5005/predict -d '{"prompt": "hihi"}'` from inside another pod in the cluster works, also the pod starts up fine which is cool, requests seem to go through (upd this is because I didn't actually use the second pod's gpus). Port 5000 should also work, probably needs some more formatting.~
  - ~ok I think I have to add another service and name it something different (eww)~
  - ~create other service when LWS is created, name it something different; have sync/streaming inference gateways predict take in a param is_lws to choose between the different service, add a "role: leader" label to the leader template~
  - ~(/workspace/model-engine/model_engine_server/infra/gateways/resources/templates inside a pod)~
  - ~`curl [2602:fb33:0:10:40ff::11]:5000/predict -d '{"args": {"prompt": "hi"}}' -H "Content-Type: application/json"` works from other pod~
  - Multinode starts up (tested with some old llama 2 7b finetune on 2 A10s)


clients/python/llmengine/data_types.py
```diff
@@ -99,6 +99,7 @@ class ModelEndpointResourceState(BaseModel):
     memory: StorageSpecificationType
     gpu_type: Optional[GpuType]
     storage: Optional[StorageSpecificationType]
+    nodes_per_worker: int = Field(..., ge=1)  # Multinode support. >1 = multinode.
```
seanshi-scale, line 102: TODO more stuff in client maybe

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -428,12 +430,28 @@ async def execute(
...
+                    model_name,
+                    framework_image_tag,
+                    endpoint_name,
+                    num_shards,
+                    quantize,
+                    checkpoint_path,
+                )
+            elif multinode:
+                raise ObjectHasInvalidValueException(
+                    f"Multinode is not supported for source {source}."
```
seanshi-scale, line 452: TODO oops framework

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1476,6 +1827,10 @@ async def _get_resources(
             deployment_config.spec.template.spec.priority_class_name == LAUNCH_HIGH_PRIORITY_CLASS
         )
 
+        # config_maps = await self._get_config_maps(
```
seanshi-scale, line 1830: why are you commented out?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1505,9 +1861,72 @@ async def _get_resources(
...
+        self,
+        endpoint_id: str,
+        deployment_name: str,
+        endpoint_type: ModelEndpointType,
+        lws_config,
+        config_maps: List,
+    ) -> ModelEndpointInfraState:
+        k8s_resource_group_name = _endpoint_id_to_k8s_resource_group_name(endpoint_id)
+
+        # config_maps = await self._get_config_maps(
```
seanshi-scale, line 1877: why are you commented out?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1505,9 +1861,72 @@ async def _get_resources(
...
+        #     endpoint_id=endpoint_id, deployment_name=k8s_resource_group_name
+        # )
+
+        # import pdb; pdb.set_trace()
```
seanshi-scale, line 1881: todo delete obvs

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1550,16 +1969,36 @@ async def _get_all_resources(
...
+                leader_worker_sets = []
+            else:
+                raise
+
         deployments_by_name = {deployment.metadata.name: deployment for deployment in deployments}
         hpas_by_name = {hpa.metadata.name: hpa for hpa in hpas}
         vpas_by_name = {vpa["metadata"]["name"]: vpa for vpa in vpas}
         keda_scaled_objects_by_name = {kso["metadata"]["name"]: kso for kso in keda_scaled_objects}
+        leader_worker_sets_by_name = {lws["metadata"]["name"]: lws for lws in leader_worker_sets}
+        # import pdb; pdb.set_trace()
```
seanshi-scale, line 1993: todo delete obvs

model-engine/model_engine_server/common/dtos/endpoint_builder.py
```diff
@@ -11,7 +11,7 @@
 )
 
 
-class BuildEndpointRequest(BaseModel):
+class BuildEndpointRequest(BaseModel):  # TODO update callsites
```
seanshi-scale, line 14: TODO check `CreateOrUpdateResourcesRequest` also

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: probably should uncomment you, iirc this was for testing

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: TODO uncomment this, it was probably for testing

model-engine/model_engine_server/infra/services/live_model_endpoint_service.py
```diff
@@ -223,7 +225,7 @@ async def get_model_endpoint(self, model_endpoint_id: str) -> Optional[ModelEndp
         # TODO we might encounter weird cache staleness issues, as the cache is potentially up
         #   to a minute behind
         model_endpoint_infra_state = await self._get_model_endpoint_infra_state(
-            record=model_endpoint_record, use_cache=True
+            record=model_endpoint_record, use_cache=False  # TODO temporarily turn off cache
```
seanshi-scale, line 228: TODO undo

model-engine/model_engine_server/infra/services/live_model_endpoint_service.py
```diff
@@ -256,7 +258,9 @@ async def list_model_endpoints(
         )
         endpoints: List[ModelEndpoint] = []
         for record in records:
-            infra_state = await self._get_model_endpoint_infra_state(record=record, use_cache=True)
+            infra_state = await self._get_model_endpoint_infra_state(
+                record=record, use_cache=False
+            )  # TODO temporarily turn off cache
```
seanshi-scale, line 263: TODO undo

clients/python/llmengine/data_types/rest.py
```diff
@@ -164,6 +165,7 @@ class CreateLLMEndpointRequest(BaseModel):
     memory: Optional[StorageSpecificationType]
     gpu_type: Optional[GpuType]
     storage: Optional[StorageSpecificationType]
+    nodes_per_worker: Optional[int] = None
```
dmchoiboi, line 168: is this something we want users to be able to provide?

model-engine/model_engine_server/common/resource_limits.py
```diff
@@ -65,7 +65,7 @@ def validate_resource_requests(
     gpus: Optional[int],
     gpu_type: Optional[GpuType],
 ) -> None:
-    """Validates whether cpu/memory requests are reasonable"""
+    """Validates whether cpu/memory requests are reasonable. Shouldn't need to validate any nodes_per_worker in the multinode case"""
```
dmchoiboi, line 68: one possible validation is nodes_per_worker must only be >1 if gpus > 8 or somethign like that. I don't imagine we want to support multinode for 1 or 2 gpus

model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
```diff
@@ -215,6 +217,47 @@ def validate_post_inference_hooks(user: User, post_inference_hooks: Optional[Lis
...
+        return
+    # can type ignore, bundle.flavor is a RunnableImageFlavor/StreamingEnhancedRunnableImageFlavor thus it has worker_command and worker_env
+    if (
+        type(bundle.flavor) in {RunnableImageFlavor, StreamingEnhancedRunnableImageFlavor}
+        and bundle.flavor.worker_command is not None  # type: ignore
+        and bundle.flavor.worker_env is not None  # type: ignore
+    ):
+        return
+    raise ObjectHasInvalidValueException(
+        f"Bundle {bundle.name} is not multinode compatible. It must be a RunnableImage and have _worker_command and _worker_args metadata."
```
dmchoiboi, line 234: was the `_` prefix intentional?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -846,32 +868,45 @@ async def create_vllm_bundle(
             final_weights_folder,
         )
 
-        vllm_cmd = f"python -m vllm_server --model {final_weights_folder} --tensor-parallel-size {num_shards} --port 5005"
+        vllm_cmd = ""
+
+        if multinode and is_leader:
+            vllm_cmd += "/workspace/init_ray.sh leader --ray_cluster_size=$RAY_CLUSTER_SIZE --own_address=$K8S_OWN_POD_NAME.$K8S_LWS_NAME.$K8S_OWN_NAMESPACE.svc.cluster.local; "
```
dmchoiboi, line 874: is there a reason this ends in a `; ` but the worker one doesn't?

dmchoiboi: oh i guess you don't start the vllm server on the worker node. Can we actually make this a separate subcmd (e.g. assign it to `ray_cmd` or something like that and rely on line 912 to add the `;`.

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -819,17 +837,21 @@ async def create_deepspeed_bundle(
...
-        user: User,
         model_name: str,
         framework_image_tag: str,
-        endpoint_unique_name: str,
         num_shards: int,
         quantize: Optional[Quantization],
         checkpoint_path: Optional[str],
         chat_template_override: Optional[str],
+        multinode: bool,
+        is_leader: bool,
```
dmchoiboi, line 849: nit: but kinda prefer if we negate this and provide an `is_worker` instead. That fits my mental model a bit better since we don't really think of leaders in the non-multinode case. We could also default is_worker to False so we don't have to pass it at all for the non multinode case

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -880,31 +915,139 @@ async def create_vllm_bundle(
...
+            num_shards,
+            quantize,
+            checkpoint_path,
+            chat_template_override,
+            multinode=True,
+            is_leader=False,
+            nodes_per_worker=nodes_per_worker,
+        )
+
+        # These env vars e.g. leader name, lws name, namespace should be filled in by Launch automatically
```
dmchoiboi, line 1012: maybe clarify that the envvars used as values are provided by Launch. wording was a bit confusing initially

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -880,31 +915,139 @@ async def create_vllm_bundle(
...
+                repository=hmi_config.vllm_repository,
+                tag=framework_image_tag,
+                command=leader_command,
+                streaming_command=leader_command,
+                protocol="http",
+                readiness_initial_delay_seconds=10,
+                healthcheck_route="/health",
+                predict_route="/predict",
+                streaming_predict_route="/stream",
+                extra_routes=[OPENAI_CHAT_COMPLETION_PATH],
```
dmchoiboi, line 1039: Add OPENAI_COMPLETION_PATH now that it's available

model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
```diff
@@ -215,6 +217,47 @@ def validate_post_inference_hooks(user: User, post_inference_hooks: Optional[Lis
...
+    """
+    Only gpu streaming endpoints can be multinode compatible.
+    """
+    if nodes_per_worker == 1:
+        return
+    if (
+        endpoint_type == ModelEndpointType.STREAMING
+        and gpu_type is not None
+        and gpus is not None
+        and gpus > 0
```
dmchoiboi, line 253: is the gpu configuration per worker or total across nodes

model-engine/model_engine_server/infra/gateways/live_model_endpoint_infra_gateway.py
```diff
@@ -151,6 +153,8 @@ async def update_model_endpoint_infra(
             gpu_type = infra_state.resource_state.gpu_type
         if storage is None:
             storage = infra_state.resource_state.storage
+        # Don't allow changing nodes_per_worker
```
dmchoiboi, line 156: should we error if nodes_per_worker is not equal to what the infra is set to?

model-engine/model_engine_server/infra/gateways/resources/live_endpoint_resource_gateway.py
```diff
@@ -59,9 +59,9 @@ async def create_or_update_resources(
             q = await self.create_queue(endpoint_record, request.build_endpoint_request.labels)
             queue_name: Optional[str] = q.queue_name
             queue_url: Optional[str] = q.queue_url
-            destination: str = q.queue_name
+            # destination: str = q.queue_name
```
dmchoiboi, line 62: delete?

clients/python/llmengine/data_types/rest.py
seanshi-scale: yes, I view it in the same vein as gpus/gpu_type/the other resource requests

model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
seanshi-scale: gpus (and all the rest of the resource requests) are per worker

model-engine/model_engine_server/infra/gateways/live_model_endpoint_infra_gateway.py
seanshi-scale: ah so I made the api such that you cannot change nodes_per_worker (changing between lws and deployment seems too annoying)

model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
seanshi-scale: oops no, this was an artifact of originally sticking stuff into bundle metadata

model-engine/model_engine_server/common/resource_limits.py
seanshi-scale: in theory yes, however I've been testing multinode on 2x1A10 nodes heh

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -507,12 +507,30 @@ async def execute(
...
+        nodes_per_worker: int,
     ) -> ModelBundle:
+        multinode = nodes_per_worker > 1
         if source == LLMSource.HUGGING_FACE:
             self.check_docker_image_exists_for_image_tag(
                 framework_image_tag, INFERENCE_FRAMEWORK_REPOSITORY[framework]
             )
-            if framework == LLMInferenceFramework.DEEPSPEED:
+            if multinode and framework == LLMInferenceFramework.VLLM:
+                bundle_id = await self.create_vllm_multinode_bundle(
```
dmchoiboi, line 518: nit: maybe move this into the framework == LLMInferenceFramework.VLLM block instead?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -846,65 +868,190 @@ async def create_vllm_bundle(
...
+            "VLLM_HOST_IP": "$(K8S_OWN_POD_NAME).$(K8S_LWS_NAME).$(K8S_OWN_NAMESPACE).svc.cluster.local",  # this needs to match what's given as --own-address in the vllm start command
+            "NCCL_SOCKET_IFNAME": "eth0",
+            "GLOO_SOCKET_IFNAME": "eth0",  # maybe don't need
+            "NCCL_DEBUG": "INFO",  # TODO remove once fully tested, will keep around for now
+            "VLLM_LOGGING_LEVEL": "INFO",  # TODO remove once fully tested, will keep around for now
+            "RAY_CLUSTER_SIZE": "$(K8S_LWS_CLUSTER_SIZE)",
+        }
+
+        leader_env = common_vllm_envs.copy()
+        worker_env = common_vllm_envs.copy()
```
dmchoiboi, line 1027: are there any difference in worker vs leader env?

model-engine/model_engine_server/domain/use_cases/sync_inference_use_cases.py
```diff
@@ -71,6 +72,14 @@ async def execute(
         await autoscaling_metrics_gateway.emit_inference_autoscaling_metric(
             endpoint_id=model_endpoint_id
         )
+        # Hack: manually resolve dns if istio is present
```
dmchoiboi, line 75: can we add a more comprehensive explanation why we do this for our future selves

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -61,6 +61,13 @@
 BASE_PATH_IN_ENDPOINT = "/app"
 
 DATADOG_ENV_VAR = {"DD_TRACE_ENABLED", "DD_SERVICE", "DD_ENV", "DD_VERSION", "DD_AGENT_HOST"}
+LWS_DEFAULT_ENV_VAR = {
+    "K8S_OWN_POD_NAME",
```
dmchoiboi, line 65: what does OWN stand for here

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -261,16 +299,59 @@ def add_datadog_env_to_main_container(deployment_template: Dict[str, Any]) -> No
...
+                "name": "K8S_OWN_NAMESPACE",
+                "valueFrom": {"fieldRef": {"fieldPath": "metadata.namespace"}},
+            },
+            {
+                "name": "K8S_LWS_NAME",
+                "valueFrom": {
+                    "fieldRef": {"fieldPath": "metadata.labels['leaderworkerset.sigs.k8s.io/name']"}
+                },
+            },
+            {  # This may not do anything actually, prefer to use LWS_LEADER_ADDRESS instead
```
dmchoiboi, line 317: is this comment still relevant?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -782,6 +1007,43 @@ async def _create_virtual_service(virtual_service: Dict[str, Any], name: str) ->
...
+                group="networking.istio.io",
+                version="v1beta1",
+                namespace=hmi_config.endpoint_namespace,
+                plural="serviceentries",
+                body=lws_service_entry,
+            )
+        except ApiException as exc:
+            if exc.status == 409:
+                logger.info(f"ServiceEntry {name} already exists, replacing")
+                # The async k8s client has a bug with patching custom objects, so we manually
```
dmchoiboi, line 1024: jank

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -846,65 +868,190 @@ async def create_vllm_bundle(
...
+            multinode=True,
+            is_worker=True,
+            nodes_per_worker=nodes_per_worker,
+        )
+
+        # These env vars e.g. K8S_OWN_POD_NAME, K8S_OWN_POD_NAME, K8S_OWN_NAMESPACE, K8S_LWS_CLUSTER_SIZE will be filled in automatically for all LWS pods through
+        # Launch's k8s_endpoint_resource_delegate
+        common_vllm_envs = {
+            "VLLM_HOST_IP": "$(K8S_OWN_POD_NAME).$(K8S_LWS_NAME).$(K8S_OWN_NAMESPACE).svc.cluster.local",  # this needs to match what's given as --own-address in the vllm start command
+            "NCCL_SOCKET_IFNAME": "eth0",
```
yunfeng-scale, line 1019: is this the correct socket to use EFA?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1146,12 +1496,43 @@ def _get_deployment_resource_name(request: CreateOrUpdateResourcesRequest) -> st
...
+        build_endpoint_request = request.build_endpoint_request
+        model_endpoint_record = build_endpoint_request.model_endpoint_record
+        flavor = model_endpoint_record.current_model_bundle.flavor
+        if isinstance(flavor, TritonEnhancedRunnableImageFlavor):
+            flavor_class = "triton-enhanced-runnable-image"
+        else:
+            flavor_class = "runnable-image"
+        if flavor_class == "triton-enhanced-runnable-image":
+            raise ValueError("LWS is not supported for Triton Enhanced Runnable Image")
+        # TODO we should really check this error earlier
```
dmchoiboi, line 1510: add the check on bundle creation? I guess regardless you'll have to handle the union type check here

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1222,23 +1627,30 @@ async def _create_or_update_resources(
...
-            sqs_queue_name=sqs_queue_name_str,
-            sqs_queue_url=sqs_queue_url_str,
-            endpoint_resource_name="pod-disruption-budget",
-        )
-        pdb_template = load_k8s_yaml("pod-disruption-budget.yaml", pdb_config_arguments)
-        await self._create_pdb(
-            pdb=pdb_template,
-            name=k8s_resource_group_name,
-        )
+        if request.build_endpoint_request.nodes_per_worker == 1:
```
dmchoiboi, line 1630: I like what you did before with assigning a "is_multinode" var to `nodes_per_worker > 1`

yunfeng-scale (commented):
thanks for the PR. put up comments so far. will come back to this later

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1332,6 +1744,54 @@ async def _create_or_update_resources(
                     destination_rule=destination_rule_template,
                     name=k8s_resource_group_name,
                 )
+        elif (
+            model_endpoint_record.endpoint_type
+            in {
+                ModelEndpointType.SYNC,
+                ModelEndpointType.STREAMING,
+            }
+            and request.build_endpoint_request.nodes_per_worker > 1
```
dmchoiboi, line 1753: Any way to clean/organize these cases? Does it make sense to nest the is_multinode check in the endpoint type check instead of making it flat?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1505,6 +1996,67 @@ async def _get_resources(
...
+        self,
+        endpoint_id: str,
+        deployment_name: str,
+        endpoint_type: ModelEndpointType,
+        lws_config,
+        config_maps: List,
+    ) -> ModelEndpointInfraState:
+        k8s_resource_group_name = _endpoint_id_to_k8s_resource_group_name(endpoint_id)
+
+        config_maps = await self._get_config_maps(
```
dmchoiboi, line 2012: we're query here again? did you mean to remove?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1390,25 +1850,54 @@ def _get_sync_autoscaling_params_from_keda(
...
+                endpoint_id=endpoint_id,
+                deployment_name=deployment_name,
+                endpoint_type=endpoint_type,
+                lws_config=lws_config,
+                config_maps=config_maps,
+            )
+        return infra_state
+
+    async def _get_resources_from_deployment_type(
+        self, endpoint_id: str, deployment_name: str, endpoint_type: ModelEndpointType, config_maps
```
dmchoiboi, line 1895: do you use the config_maps here anywhere?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1476,6 +1962,10 @@ async def _get_resources(
             deployment_config.spec.template.spec.priority_class_name == LAUNCH_HIGH_PRIORITY_CLASS
         )
 
+        config_maps = await self._get_config_maps(
```
dmchoiboi, line 1965: remove and use the value passed in?

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1634,9 +2205,27 @@ async def _get_all_resources(
...
+                is_key_an_endpoint_id,
+                await self._get_resources_from_lws_type(
+                    endpoint_id, deployment_name, endpoint_type, lws_config, all_config_maps
+                ),
+            )
         return infra_states
 
     async def _delete_resources_async(self, endpoint_id: str, deployment_name: str) -> bool:
+
+        # TODO multinode if we ever decide to support that
```
dmchoiboi, line 2227: what do you mean by this? it seems you added delete_lws here? what else needs to be added for multinode support?

model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
dmchoiboi: I'm realizing that the term `worker` is slightly overloaded and imprecise since it's defined and used in both `nodes_per_worker` (where it represents the multinode entity) and in `leaderworkerset` (where it represents an individual node).

Could you update our documentation so it's clear what exactly these resources are tied to. 



So

dmchoiboi (approved):
Looks good overall! just a few comments

clients/python/llmengine/model.py
```diff
@@ -110,6 +111,13 @@ def create(
             gpus (`Optional[int]`):
                 Number of gpus each worker should get, e.g. 0, 1, etc. Can be inferred from the model size.
 
+            nodes_per_worker (`int`):
+                Number of nodes per worker. Used to request multinode serving. This must be greater than or equal to 1.
+                Controls how many nodes to dedicate to one instance of the model.
+                Any compute resource requests (i.e. cpus, memory, storage) apply to each node, thus the total resources
+                allocated are multiplied by this number. This is useful for models that require more memory than a single node can provide.
+                Note: autoscaling is not supported for multinode serving.
```
yunfeng-scale, line 119: we'd add it in the future? since LWS does support HPA

yunfeng-scale: is it because we don't have istio metrics? we do have vLLM metrics

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: yeah it is, we should be using LWS_LEADER_ADDRESS in the vllm bundle

dmchoiboi: should we remove the env var then?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
seanshi-scale: there aren't, this is likely an artifact of a long time ago where I thought there was a chance the env values could be different

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: it's basically synonymous with "SELF", i.e. the value of that env var is the pod's own name, e.g. in a pod named `pod1` the value of the env var is `pod1`

clients/python/llmengine/model.py
seanshi-scale: I didn't add HPA to keep things simple here; I want to be more certain that the LWS replicas can start up properly (I found issues with starting up LWS in the past with random NCCL errors that sometimes happened). Also, I don't think our current use case would require autoscaling. We can certainly add autoscaling to LWS if we need though. (also yeah we probably don't have istio metrics)

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -782,6 +1007,43 @@ async def _create_virtual_service(virtual_service: Dict[str, Any], name: str) ->
                 logger.exception("Got an exception when trying to apply the VirtualService")
                 raise
 
+    @staticmethod
+    async def _create_lws_service_entry(lws_service_entry: Dict[str, Any], name: str) -> None:
```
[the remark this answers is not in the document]
seanshi-scale: Empirically, without the service entry, we're not able to make requests to the Service's IP. The service entry yaml template is specific to LWS in the sense that it has a few parameters set specifically for the LWS case, e.g. 
```
location: MESH_EXTERNAL
resolution: NONE
```

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -329,6 +410,18 @@ def _get_env_value_from_envlist(
                 return envvar.value
         return None
 
+    @staticmethod
+    def _get_env_value_from_envlist_for_lws(
```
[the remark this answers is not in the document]
seanshi-scale: changed to _for_custom_object, since it applies to those specifically (i.e. not k8s Deployment/Pod/Job/Service/etc.)

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: yeah rip, it is decently annoying

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
seanshi-scale: got rid of leader/worker_env anyways lol

clients/python/llmengine/data_types/rest.py
seanshi-scale: we do have a route to figure out this number for users if they don't set any resource requests (through the "default resource requests" behavior), e.g. someone creating a 405b unquantized without specifying any resource requests will default to multinode. Otherwise I think the flexibility/explicitness is probably good?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
seanshi-scale: IIRC I copied this from the setup on the training side of things

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: Currently multinode async endpoints are not supported, since we don't actually have that template generated. I'd guess that we only need to do the `delete_lws()` so _delete_resources_async probably would work, but I'd want to only mark it as done once async endpoints have been e2e implemented

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -261,16 +299,59 @@ def add_datadog_env_to_main_container(deployment_template: Dict[str, Any]) -> No
     user_container["env"] = user_container_envs
 
 
+def add_lws_default_env_vars_to_container(container: Dict[str, Any]) -> None:
+    container_envs = []
+    container_envs.extend(
```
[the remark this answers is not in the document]
seanshi-scale: This continues the pattern in the deployment case where we fill in the env vars in code

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1550,16 +2102,34 @@ async def _get_all_resources(
...
         vpas_by_name = {vpa["metadata"]["name"]: vpa for vpa in vpas}
         keda_scaled_objects_by_name = {kso["metadata"]["name"]: kso for kso in keda_scaled_objects}
+        leader_worker_sets_by_name = {lws["metadata"]["name"]: lws for lws in leader_worker_sets}
         all_config_maps = await self._get_all_config_maps()
         # can safely assume hpa with same name as deployment corresponds to the same Launch Endpoint
         logger.info(f"Orphaned hpas: {set(hpas_by_name).difference(set(deployments_by_name))}")
         logger.info(f"Orphaned vpas: {set(vpas_by_name).difference(set(deployments_by_name))}")
         infra_states = {}
-        logger.info(f"Got data for {list(deployments_by_name.keys())}")
+        logger.info(
```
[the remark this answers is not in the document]
seanshi-scale: I feel like we probably don't need to? The "constraint" that the two lists should specify is that their intersection is empty and their union is all endpoints (modulo entries in the db that don't point to anything or something like that), in that sense I don't think we care about the difference.

In any case I don't really mind either way, but the same line feels technically feels better to me

model-engine/model_engine_server/domain/use_cases/model_endpoint_use_cases.py
seanshi-scale: Added some docs to the llm engine client side of things

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -391,6 +484,63 @@ def _get_common_endpoint_params(self, deployment_config: V1Deployment) -> Common
...
+        gpus = int((resources["limits"] or dict()).get("nvidia.com/gpu", 0))
+        storage = resources["requests"].get("ephemeral-storage")
+
+        envlist = launch_container["env"]
+        # There really isn't a bundle_url for LWS since those use RunnableImages
+        bundle_url = self._get_env_value_from_envlist_for_lws(envlist, "BUNDLE_URL") or image
+        aws_role = self._get_env_value_from_envlist_for_lws(envlist, "AWS_PROFILE")
+        results_s3_bucket = self._get_env_value_from_envlist_for_lws(envlist, "RESULTS_S3_BUCKET")
+
+        # Temporary fix: new LIRA endpoints created should have these env vars
```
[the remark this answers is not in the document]
seanshi-scale: it's not, it shouldn't hurt to do that replace though heh

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: oh lmao oops the check actually does happen on endpoint create (e.g. `validate_bundle_multinode_compatibility()`)

clients/python/llmengine/data_types/rest.py
```diff
@@ -86,13 +86,18 @@ class ModelEndpointDeploymentState(BaseModel):
 class ModelEndpointResourceState(BaseModel):
     """
     This is the entity-layer class for the resource settings per worker of a Model Endpoint.
+    Note: the values for cpus/gpus/memory/storage are per node, i.e. a single "worker" may consist of
```
dmchoiboi, line 89: thx

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: I feel like it's probably nicer to keep things flat, since there are some commonly created resources (and it's nice to group things by function)

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: yes oops I had commented it out, and thought I needed it and uncommented it by mistake

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
seanshi-scale: oops fixed

PR #578, opened 2024-07-26
Shared pydantic configs

dmchoiboi:
# Pull Request Summary

Extract custom base configs to a custom pydantic base model. In particular, we want to remove the `protected_namespaces` to remove all the warning logs around that

## Test Plan and Usage Guide


model-engine/model_engine_server/common/pydantic_types/endpoint_predict_payload.py
```diff
@@ -1,10 +0,0 @@
-from typing import Any, Optional
-
-from pydantic import BaseModel
-
-
-class EndpointPredictPayload(BaseModel):
```
seanshi-scale, line 6: is this not used anywhere?

seanshi-scale (approved):

model-engine/model_engine_server/common/pydantic_types/endpoint_predict_payload.py
dmchoiboi: I did a ctrl-f and did not see it anywhere. We'll see if tests fail..

PR #579, opened 2024-07-26
Add autogenerated openai spec

dmchoiboi:
# Pull Request Summary

Autogenerate openai pydantic models to use as base models

## Test Plan and Usage Guide


seanshi-scale (approved):
looks like there's some things in there we won't need (e.g. specific models) but this seems fine
if the upstream spec changes, will we rewrite apis/code to adapt ASAP, or on some cadence?

dmchoiboi:
> if the upstream spec changes, will we rewrite apis/code to adapt ASAP, or on some cadence?

I'll write some wrapper classes around these models for our internal usages. We can either update on a cadence or only when relevant changes need to be ingested

PR #581, opened 2024-07-29
Make configs backwards-compatible

squeakymouse:
# Pull Request Summary

Filter out extra config fields to fix `TypeError: __init__() got an unexpected keyword argument 'cache_redis_url'` error in customer VPC occurring from `cache_redis_url` being an old parameter that some versions of the model engine instantiation code still pass in

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/common/config.py
+++ b/model-engine/model_engine_server/common/config.py
@@ -1,6 +1,7 @@
 # Keep in line with service_config_{*}.yaml
 # This file loads sensitive data that shouldn't make it to inference docker images
 # Do not include this file in our inference/endpoint code
+import inspect
 import os
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


dmchoiboi:
Can you add more details on what the error is here? It's not clear to me what this is fixing

model-engine/model_engine_server/common/config.py
```diff
@@ -80,6 +80,11 @@ class HostedModelInferenceServiceConfig:
     def from_yaml(cls, yaml_path):
         with open(yaml_path, "r") as f:
             raw_data = yaml.safe_load(f)
+        raw_data = {
```
dmchoiboi, line 83: Can we extract this into a from_dict classmethod and use that instead?

e.g.
```
@classmethod
def from_json(cls, json):
    return cls(**{
        k: v for k, v in json.items() 
        if k in inspect.signature(cls).parameters
    })
```

dmchoiboi (approved):

PR #586, opened 2024-08-02
Enable passing in headers through the client

dmchoiboi:
# Pull Request Summary

This enables custom auth header implementations

## Test Plan and Usage Guide

### DIFF
--- a/clients/python/llmengine/api_engine.py
+++ b/clients/python/llmengine/api_engine.py
@@ -61,7 +61,7 @@ class APIEngine:
 
     @classmethod
     def _get(
-        cls, resource_name: str, timeout: int, headers: Optional[Dict[str, Any]] = None
+        cls, resource_name: str, timeout: int, headers: Optional[Dict[str, str]] = None
     ) -> Dict[str, Any]:
         base_path = get_base_path()
         api_key = get_api_key()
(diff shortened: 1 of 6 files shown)


clients/python/llmengine/api_engine.py
```diff
@@ -60,13 +60,15 @@ def validate_api_key(cls):
             )
 
     @classmethod
-    def _get(cls, resource_name: str, timeout: int) -> Dict[str, Any]:
+    def _get(
+        cls, resource_name: str, timeout: int, headers: Optional[Dict[str, Any]] = None
```
seanshi-scale, line 64: nit: should headers be Optional[Dict[str, str]]? https://requests.readthedocs.io/en/latest/user/quickstart/#custom-headers mentions values need to be string/bytestring/unicode

clients/python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "scale-llm-engine"
-version = "0.0.0.beta34"
+version = "0.0.0.beta35"
```
seanshi-scale, line 3: there should be two other places where this version bump needs to happen (for a total of 3 places)

seanshi-scale (commented):
had a comment about version bump (there's three places to put the version num), also a nit, otherwise looks good

seanshi-scale (approved):

PR #595, opened 2024-08-14
Get Gemma2 working

dmchoiboi:
# Pull Request Summary

* Mostly consolidate vllm vs vllm_batch image generation - use same dockerfile w/ multistage targets
* Install flashinfer 0.0.8 for the corresponding torch version (2.3)
* Clean up vllm_server to use vllm's official openai server (plus some added debugging + backwards-compatible routes)

## Test Plan and Usage Guide


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -797,6 +798,9 @@ async def create_vllm_bundle(
         if "llama-3-70b" in model_name:
             subcommands[-1] = subcommands[-1] + " --gpu-memory-utilization 0.95 --enforce-eager"
 
+        if "gemma-2" in model_name:
```
seanshi-scale, line 801: slightly weird we don't use `infer_addition_engine_args_from_model_name` here, and it looks like they do slightly different things, i.e. for gpu mem util matching llama-3-70b exactly or anything with >=70b params, but I guess that was the case to begin with

model-engine/model_engine_server/inference/vllm/Dockerfile.vllm
```diff
@@ -0,0 +1,48 @@
...
+
+# symlink python to python3
+RUN ln -s /usr/bin/python3 /usr/bin/python
+
+FROM base AS vllm
+
+WORKDIR /workspace
+
+COPY model-engine/model_engine_server/inference/vllm/requirements.txt /workspace/requirements.txt
+RUN pip install -r requirements.txt
```
seanshi-scale, line 29: do we still end up needing the pip install requirements.txt? they're just vllm + pydantic at this point, so if we drop the vllm req from requirements then (as long as the pydantic install doesn't bust the python env) we can just omit having to make sure the VLLM_VERSION in the build script + vllm version in req.txt match

model-engine/model_engine_server/inference/vllm/vllm_server.py
```diff
@@ -1,89 +1,40 @@
...
-
-
-@app.get("/v1/models")
-async def show_available_models():
-    models = await openai_serving_chat.show_available_models()
-    return JSONResponse(content=models.model_dump())
-
-
-@app.post("/v1/chat/completions")
-async def create_chat_completion(request: OpenAIChatCompletionRequest, raw_request: Request):
```
seanshi-scale, line 63: guess we're gonna rethink these?

seanshi-scale (commented):
had a q about the build process, wondering if we can get rid of the vllm in req.txt (since ostensibly that should be in the base image?)

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
dmchoiboi: Agreed; I think there is room for consolidation/improvement but don't want to make too many changes in this PR

model-engine/model_engine_server/inference/vllm/Dockerfile.vllm
dmchoiboi: true, i could remove it from the dockerfile; still want it around for local development

model-engine/model_engine_server/inference/vllm/vllm_server.py
dmchoiboi: I use `init_app` from vllm's openai entrypoint package which define these routes for us. I just added our /predict /stream routes

seanshi-scale: oh got it, that makes sense

seanshi-scale (approved):

PR #606, opened 2024-08-22
Relax pydantic constraint for client

dmchoiboi:
# Pull Request Summary

Upping pydantic requirement is proving problematic when rolling out. Will relax the requirement and autogenerate v1 compatible objects instead

Also, add client methods for getting and cancelling batch completions 

## Test Plan and Usage Guide


yunfeng-scale:
what's the error when rollout?

dmchoiboi:
> what's the error when rollout?

Was testing the prior client out with some users and we ran into issues w/ some of their code still being on Pydantic v1. To facilitate migration, I decided to just update the client to also be pydantic v1 compatible, even though we really should move everyone off

yunfeng-scale (approved):

yunfeng-scale:
> > what's the error when rollout?
> 
> Was testing the prior client out with some users and we ran into issues w/ some of their code still being on Pydantic v1. To facilitate migration, I decided to just update the client to also be pydantic v1 compatible, even though we really should move everyone off

how hard is it for them to migrate?

dmchoiboi:
> how hard is it for them to migrate?

I don't know; but I don't want to put myself in a position where people aren't using this new API cause they're blocked on migrating

clients/python/llmengine/completion.py
```diff
@@ -507,9 +509,15 @@ def batch_create(
             data_parallelism (int):
                 The number of parallel jobs to run. Data will be evenly distributed to the jobs. Defaults to 1.
 
+            priority (str):
```
Georgepu1, line 512: nit: comment for `labels`

PR #608, opened 2024-09-04
Http Forwarder updates

dmchoiboi:
# Pull Request Summary

Update http forwarder to enable forwarding to additional routes to downstream services

## Test Plan and Usage Guide


model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -36,14 +39,18 @@ def get_config():
     )
 
 
-def get_forwarder_loader():
+def get_forwarder_loader(destination_path: Optional[str] = None):
     config = get_config()
+    if destination_path:
```
seanshi-scale, line 44: wait does this do anything? since we look at `config['sync']` instead of `config['predict_route']`

dmchoiboi: ah typo

seanshi-scale: also get rid of the ["sync"] in `config["sync"]["predict_route"] = destination_path`?

model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
```diff
@@ -9,6 +9,7 @@ forwarder:
     model_engine_unwrap: true
     serialize_results_as_string: true
     forward_http_status: true
+    extra_routes: []
```
seanshi-scale, line 12: btw it might be hard to define behavior from here, we may need to end up passing this as some runtime parameter to the http forwarder (e.g. through cli args)

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -111,6 +118,47 @@ async def event_generator():
...
+# NOTE: it is important for this to be defined AFTER the /predict and /stream endpoints
+# because FastAPI will match the first route that matches the request path
+async def predict_or_stream(
+    request: EndpointPredictV1Request,
+    background_tasks: BackgroundTasks,
+    sync_forwarder: Optional[Forwarder],
+    stream_forwarder: Optional[StreamingForwarder],
+    limiter=Depends(get_concurrency_limiter),
+):
+    if stream_forwarder and request.args and request.args.root.get("stream", False):
```
dmchoiboi, line 132: if we don't like having this implicit expectation of having "stream" in the body, we can separate out a prefix for sync vs stream

e.g. 
/stream/{path} -> /{path} 
/sync/{path} -> /{path}

The tradeoff is that the gateway would have to know to prefix stream requests with /stream and sync with /sync. Both options have the potential to cause confusion in some way

model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
dmchoiboi: yeah i added this here to illustrate that the field exists. We'd probably need to do something similar to how we pass in the predict route as env var at model bundle creation time

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -111,6 +118,47 @@ async def event_generator():
...
+def add_extra_routes():
+    """Read extra_routes from config and dynamically add routes to app"""
+    config = get_config()
+    # aggregate a list of routes -> (sync, stream) support
+    extra_routes: Dict[str, Tuple[Optional[Forwarder], Optional[StreamingForwarder]]] = dict()
+    for route in config.get("sync", {}).get("extra_routes", []):
+        extra_routes[route] = (load_forwarder(route), None)
+    for route in config.get("stream", {}).get("extra_routes", []):
+        extra_routes[route] = (
+            extra_routes.get(route, (None, None))[0],
```
seanshi-scale, line 149: is this so that some routes are marked as both sync and streaming enabled?

dmchoiboi: yeah.. e.g. if we use vllm openai server out of box, they use the same routes for sync and stream and use body:stream = true to distinguish requests

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -111,6 +118,47 @@ async def event_generator():
...
+        return predict(request, background_tasks, sync_forwarder, limiter)
+    else:
+        raise Exception("No forwarder configured for this route")
+
+
+def add_extra_routes():
+    """Read extra_routes from config and dynamically add routes to app"""
+    config = get_config()
+    # aggregate a list of routes -> (sync, stream) support
+    extra_routes: Dict[str, Tuple[Optional[Forwarder], Optional[StreamingForwarder]]] = dict()
```
yunfeng-scale, line 144: possible to simplify L144-L151? i think you can define two dicts instead of one to make the logic more clear

yunfeng-scale (commented):
can you post the design doc first?

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
dmchoiboi: true; sorry will clean this up

model-engine/model_engine_server/inference/forwarding/forwarding.py
```diff
@@ -509,13 +509,22 @@ def _substitute_config_overrides(config: dict, config_overrides: List[str]) -> N
             raise ValueError(f"Error setting {key_path} to {value} in {config}") from e
 
 
+def _cast_value(value: Any) -> Any:
```
seanshi-scale, line 512: does json.loads() work here or do we need something different? seems we're parsing nested python lists of integers

dmchoiboi: it's not quite json.loads cause the values themselves don't need to be wrapped in strings. Right now i'm just trying to support ARGS=[/v1/chat/completions]

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -111,43 +113,133 @@ async def event_generator():
...
-        str(args.graceful_timeout),
-        *envs,
-        "model_engine_server.inference.forwarding.http_forwarder:app",
-        *extra_args,
-    ]
-    subprocess.run(command)
+        os.environ["CONFIG_OVERRIDES"] = ";".join(args.set)
+
+    asyncio.run(
+        run_server(
```
seanshi-scale, line 235: are we getting rid of gunicorn here? probably fine, but want to make sure this doesn't kill our ability to handle lots of concurrent requests or anything like that

dmchoiboi: yeah i'll do some rough tests to make sure this is ok

yunfeng-scale: thanks for doing this!

yunfeng-scale (approved):

PR #613, opened 2024-09-16
Option to skip AWS profile set

seanshi-scale:
# Pull Request Summary

Allow skipping setting AWS_PROFILE, if we're in environments that handle setting AWS_PROFILE for us

## Test Plan and Usage Guide

will run a job on such an environment, if it works then I will be happy and merge this.

### DIFF
--- a/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
+++ b/model-engine/model_engine_server/inference/batch_inference/vllm_batch.py
@@ -31,7 +31,9 @@ CONFIG_FILE = os.getenv("CONFIG_FILE")
 AWS_REGION = os.getenv("AWS_REGION", "us-west-2")
 MODEL_WEIGHTS_FOLDER = os.getenv("MODEL_WEIGHTS_FOLDER", "./model_weights")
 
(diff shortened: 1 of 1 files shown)


yunfeng-scale (commented):
hmm, what about adding a cloud provider env var?

seanshi-scale:
> hmm, what about adding a cloud provider env var?

I think that would probably make sense in the case we need to support this on "solely Azure" or "solely GCP", but right now I don't think a cloud provider env var would be granular enough for our use case (we need to run compute on one cloud and use the blob storage of another cloud for instance)

yunfeng-scale:
> > hmm, what about adding a cloud provider env var?
> 
> I think that would probably make sense in the case we need to support this on "solely Azure" or "solely GCP", but right now I don't think a cloud provider env var would be granular enough for our use case (we need to run compute on one cloud and use the blob storage of another cloud for instance)

i see. what about a more granular env var? i feel SKIP_XXX is not conveying what we're trying to do

yunfeng-scale (approved):
will see if this code needs change when running in other cloud providers

PR #617, opened 2024-09-23
Enable users to force redeploy endpoints

dmchoiboi:
# Pull Request Summary

This allows users to redeploy endpoints to pick up changes in llm engine implementation in bundle creation even when the underlying endpoint spec hasn't changed

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/common/dtos/llms/model_endpoints.py
+++ b/model-engine/model_engine_server/common/dtos/llms/model_endpoints.py
@@ -149,6 +149,14 @@ class UpdateLLMModelEndpointV1Request(BaseModel):
         description="A Jinja template to use for this endpoint. If not provided, will use the chat template from the checkpoint",
     )
 
+    force_redeploy: Optional[bool] = False
+    """
+    Whether to force redeploy the endpoint.
+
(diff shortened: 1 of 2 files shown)


model-engine/model_engine_server/common/dtos/llms/model_endpoints.py
```diff
@@ -149,6 +149,14 @@ class UpdateLLMModelEndpointV1Request(BaseModel):
         description="A Jinja template to use for this endpoint. If not provided, will use the chat template from the checkpoint",
     )
 
+    force_redeploy: Optional[bool] = False
+    """
+    Whether to force redeploy the endpoint.
+
+    If True, the endpoint will be redeployed even if the update request does not change the endpoint spec.
+    This enables users to redeploy an endpoint if there are some implementation changes that we'd like to apply to the endpoint
```
seanshi-scale, line 157: this forces creation of a new bundle right? IIUC we need this if we're changing how the bundle works, but not if we are only changing how the forwarder works?

if that's the case, could we mention it in the docstring?

seanshi-scale (approved):
had a nit about some documentation, lgtm after that though

PR #621, opened 2024-09-24
Fix batch compeltion v2 for oai completion

dmchoiboi:
# Pull Request Summary

vllm oai completion requires raw request, which we don't have for batch completion. hack it by instantiating a dummy request.

## Test Plan and Usage Guide


yunfeng-scale (approved):
lol okay

PR #627, opened 2024-09-30
Add OpenAI compatible v2 completion

dmchoiboi:
# Pull Request Summary

* Add openai compatible APIs for legacy completions
* HTTP forwarder improvements
  * Fix route registration logic - dependency injection wasn't working
  * Add async methods for predict forwarding - this was causing major bottlenecks 
  * Make post inference handler optional - makes testing locally easier

## Test Plan and Usage Guide


model-engine/model_engine_server/inference/forwarding/forwarding.py
```diff
@@ -142,7 +148,49 @@ class Forwarder(ModelEngineSerializationMixin):
     serialize_results_as_string: bool
     wrap_response: bool
     forward_http_status: bool
-    post_inference_hooks_handler: PostInferenceHooksHandler
+    post_inference_hooks_handler: Optional[PostInferenceHooksHandler] = None
+
+    async def forward(self, json_payload: Any) -> Any:
```
dmchoiboi, line 153: This is a async version of __call__. I didn't want to refactor all uses of __call__ so i'm just creating a separate function to do this

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -181,23 +182,30 @@ def add_extra_routes(app: FastAPI):
...
+                stream_forwarder=Depends(get_stream_forwarder),
                 limiter=Depends(get_concurrency_limiter),
             ):
                 if not request.args:
                     raise Exception("Request has no args")
                 if request.args.root.get("stream", False) and stream_forwarder:
                     return await stream(request, stream_forwarder, limiter)
                 elif request.args.root.get("stream") is not True and sync_forwarder:
-                    return predict(request, background_tasks, sync_forwarder, limiter)
+                    return await predict(request, background_tasks, sync_forwarder, limiter)
```
dmchoiboi, line 208: this was a big woopsie; this is a blocking method in an async function which essentially blocks the thread and severely restricting concurrency

model-engine/model_engine_server/inference/forwarding/forwarding.py
yunfeng-scale: rename to `aforward` or `async_forward` to distinguish?

model-engine/model_engine_server/inference/forwarding/forwarding.py
```diff
@@ -142,7 +149,49 @@ class Forwarder(ModelEngineSerializationMixin):
...
+
+        if self.wrap_response:
+            response = self.get_response_payload(using_serialize_results_as_string, response)
+
+        if self.forward_http_status:
+            return JSONResponse(content=response, status_code=response_raw.status)
+        else:
+            return response
 
     def __call__(self, json_payload: Any) -> Any:
```
yunfeng-scale, line 196: we think sync post causes health check failures at high load?

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -73,18 +73,19 @@ def load_streaming_forwarder(destination_path: Optional[str] = None):
...
     forwarder=Depends(load_forwarder),
     limiter=Depends(get_concurrency_limiter),
 ):
     with limiter:
         try:
-            response = forwarder(request.model_dump())
-            background_tasks.add_task(
-                forwarder.post_inference_hooks_handler.handle, request, response
-            )
+            response = await forwarder.forward(request.model_dump())
```
yunfeng-scale, line 84: what are the other places that use __call__?

yunfeng-scale (approved):

model-engine/model_engine_server/inference/forwarding/forwarding.py
dmchoiboi: eh, is prefixing w/ `a` standard practice? I feel like we should just default to async

yunfeng-scale: not really, mostly because there's already one function that's sync, and i want to make it clear about the differences

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -2436,9 +2444,260 @@ async def _response_chunk_generator(
...
+        await autoscaling_metrics_gateway.emit_inference_autoscaling_metric(
+            endpoint_id=model_endpoint.record.id
+        )
+        endpoint_content = _model_endpoint_entity_to_get_llm_model_endpoint_response(model_endpoint)
+
+        validate_endpoint_supports_openai_completion(model_endpoint, endpoint_content)
+
+        # if inference framework is VLLM, we need to set the model to use the weights folder
+        if endpoint_content.inference_framework == LLMInferenceFramework.VLLM:
+            request.model = VLLM_MODEL_WEIGHTS_FOLDER
```
seanshi-scale, line 2544: does this mean we can't change the model weights folder going forward?

dmchoiboi: hm, not necessarily.. maybe we can think of a better way to do this. I think on the vllm side, we could actually provide a served_model_name or something so you can actually provide the model name instead of the model directory name

seanshi-scale (approved):

PR #634, opened 2024-10-14
0-N scaling

seanshi-scale:
# Pull Request Summary

## Test Plan and Usage Guide

### DIFF
--- a/charts/model-engine/templates/service_template_config_map.yaml
+++ b/charts/model-engine/templates/service_template_config_map.yaml
@@ -472,6 +472,7 @@ data:
           unsafeSsl: "false"
           databaseIndex: "${REDIS_DB_INDEX}"
       {{- end }}
+      # TODO
   {{- range $device := tuple "gpu" }}
   {{- range $mode := tuple "streaming"}}
   leader-worker-set-{{ $mode }}-{{ $device }}.yaml: |-


model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1283,6 +1283,7 @@ async def execute(
             min_workers=request.min_workers,
             max_workers=request.max_workers,
             endpoint_type=request.endpoint_type,
+            sync_can_scale_from_zero=self.model_endpoint_service.can_autoscale_sync_stream_endpoints_from_zero(),
```
dmchoiboi, line 1286: why sync only?

seanshi-scale: I can rename this if you want, I think my thinking was "the variable name is getting kinda long"

dmchoiboi: maybe we just remove `sync` and `stream` in these? and just have like a comment/disclaimer that async isn't implemented/supported?

seanshi-scale: async already scales from zero (that was one of its advantages before)

dmchoiboi: oh i guess that makes sense. I think we can still remove the sync/stream from the variable name

seanshi-scale: wait actually think "http_scale_from_zero" seems reasonable?

dmchoiboi: lol until we have to support websockets XP

i'm fine with it though

model-engine/model_engine_server/domain/use_cases/batch_job_use_cases.py
```diff
@@ -68,6 +68,7 @@ async def execute(
             min_workers=0,
             max_workers=request.resource_requests.max_workers,
             endpoint_type=ModelEndpointType.ASYNC,
+            http_can_scale_from_zero=self.model_endpoint_service.can_scale_http_endpoint_from_zero(),
```
dmchoiboi, line 71: can we name this consistently with model_endpoint_service?

dmchoiboi (approved):

charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -472,6 +472,12 @@ data:
           unsafeSsl: "false"
           databaseIndex: "${REDIS_DB_INDEX}"
       {{- end }}
+      - type: prometheus
```
yunfeng-scale, line 475: curious how redis and istio metrics work together here?

seanshi-scale: it'll take whichever metric gives the higher value, in practice the redis metric only gives a fast 0-1 signal (since the length of the redis queue is at most 1 barring weird race conditions) and the prometheus metric (which is identical to the 1-N hpa scaling metric) handles the 1-N condition

PR #638, opened 2024-10-15
Updates to get tool use working

dmchoiboi:
* Update oai spec to remove strict flag default to workaround vllm incompatibilty
* additional flags to set through API
* fix handling vllm args false flags
# Pull Request Summary

## Test Plan and Usage Guide

Created llama-3-1-8b-instruct endpoint


model-engine/model_engine_server/common/dtos/llms/vllm.py
```diff
@@ -69,6 +69,16 @@ class VLLMModelConfig(BaseModel):
         description="A Jinja template to use for this endpoint. If not provided, will use the chat template from the checkpoint",
     )
 
+    tool_call_parser: Optional[str] = Field(
+        None,
+        description="Tool call parser",
```
yunfeng-scale, line 74: what's tool call parser?

yunfeng-scale (approved):

model-engine/model_engine_server/common/dtos/llms/vllm.py
dmchoiboi: see https://github.com/vllm-project/vllm/blob/5d264f4ab8d008f0ac5b7f0adb7189d70136f3ec/docs/source/serving/openai_compatible_server.md#tool-calling-in-the-chat-completion-api-1 for documentation

seanshi-scale (approved):

PR #641, opened 2024-10-16
Remove model name override

dmchoiboi:
# Pull Request Summary

After deploying https://github.com/scaleapi/llm-engine/pull/639 to all endpoints, we can remove the model.name override we do here

## Test Plan and Usage Guide


seanshi-scale (approved):
would we need to redeploy everything on llm engine? (including people's private models and all of that)

dmchoiboi:
> would we need to redeploy everything on llm engine? (including people's private models and all of that)
This would only apply for endpoints that support oia compatible APIs

PR #645, opened 2024-10-17
Miscellaneous improvments

dmchoiboi:
# Pull Request Summary

* Fix client usage of 'Annotated' type
* Add request id to batch completions API failures

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/api/v2/batch_completion.py
+++ b/model-engine/model_engine_server/api/v2/batch_completion.py
@@ -14,7 +14,12 @@ from model_engine_server.common.dtos.llms.batch_completion import (
     UpdateBatchCompletionsV2Response,
 )
 from model_engine_server.core.auth.authentication_repository import User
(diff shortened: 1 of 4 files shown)


seanshi-scale (approved):

model-engine/model_engine_server/api/v2/batch_completion.py
```diff
@@ -149,5 +156,5 @@ async def cancel_batch_completion(
         )
         raise HTTPException(
             status_code=500,
-            detail="Internal server error",
+            detail=f"Internal server error. request_id: {request_id}",
```
yunfeng-scale, line 159: tbh, we should update all of these `Internal server error` to be explicit about what exact error it is. there's no need to hide it from users

dmchoiboi: yes i agree.. i don't want to address this right now though

PR #647, opened 2024-10-23
[WIP] Add Templating for Configmaps

ewkoch:
# Pull Request Summary

Add templating for configmaps so that they don't have to be built separately. Aiming to comment a default `values.yaml` file so we can clearly show what needs to be configured for working deployment.

## Test Plan and Usage Guide

Need to test/deploy from scratch prior to merge.


charts/model-engine/templates/_helpers.tpl
```diff
@@ -345,6 +345,13 @@ env:
   - name: REDIS_AUTH_TOKEN
     value: {{ .Values.redis.auth }}
   {{- end }}
+  {{- if .Values.redis.authSecret }}
+  - name: REDIS_AUTH_TOKEN
+    valueFrom:
+      secretKeyRef:
+        name: {{ .Values.redis.authSecret }}
+        key: auth_token
+  {{- end }}
```
ewkoch, line 354: Need refactor/replace above, this should come from secret. Also need to verify this is actually used. Vaguely remember seeing that it was unnecessary

charts/model-engine/templates/aws_config_map.yaml
```diff
@@ -20,6 +20,9 @@ data:
     [profile {{ $profileName }}]
     role_arn = {{ index $annotations "eks.amazonaws.com/role-arn" }}
     web_identity_token_file = /var/run/secrets/eks.amazonaws.com/serviceaccount/token
+    [profile {{ $.Values.serviceAccount.sqsProfileName }}]
+    role_arn = {{ index $annotations "eks.amazonaws.com/role-arn" }}
+    web_identity_token_file = /var/run/secrets/eks.amazonaws.com/serviceaccount/token
```
ewkoch, line 25: Ideally consolidate this to one profile- need to verify that the SQS profile  is assumable / check whether it has to match $profileName

charts/model-engine/templates/inference_framework_config.yaml
```diff
@@ -2,17 +2,18 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ include "modelEngine.fullname" . }}-inference-framework-latest-config
+  namespace: {{ .Release.Namespace }}
   labels:
     product: common
     team: infra
   annotations:
-    "helm.sh/hook": pre-install
+    "helm.sh/hook": pre-install,pre-upgrade
```
ewkoch, line 10: without this you can't update the vllm tags on upgrades

charts/model-engine/templates/inference_framework_config.yaml
```diff
@@ -2,17 +2,18 @@ apiVersion: v1
...
     "helm.sh/hook-weight": "-2"
 data:
   deepspeed: "latest"
   text_generation_inference: "latest"
-  vllm: "latest"
-  vllm_batch: "latest"
-  vllm_batch_v2: "latest"
+  vllm: "{{ .Values.vllmTag }}"
+  vllm_batch:  "{{ .Values.vllmTag }}"
+  vllm_batch_v2:  "{{ .Values.vllmTag }}"
```
ewkoch, line 17: nice to specify

charts/model-engine/templates/istio-virtualservice.yaml
```diff
@@ -12,13 +12,9 @@ metadata:
...
   hosts:
-  {{- range .Values.virtualservice.hostDomains }}
-    - "{{ $fullName }}.{{ . }}"
-  {{- end }}
+    - model-engine.{{ $.Values.global.networking.internalDomain }}
   gateways:
-  {{- range .Values.virtualservice.gateways }}
-    - {{ . | quote }}
-  {{- end }}
+    - {{ $.Values.global.networking.internalGateway }}
```
ewkoch, line 17: im open to scrapping this- this was part of a DNS refactor we did to make specification of virtual services simpler

charts/model-engine/templates/service_config_map.yaml
```diff
@@ -11,46 +12,110 @@ metadata:
...
-    {{- with .Values.config.values.launch }}
-    {{- range $key, $value := . }}
-    {{ $key }}: {{ $value | quote }}
-    {{- end }}
-    {{- end }}
-  infra_service_config: |-
-    env: {{ .Values.context | quote }}
-    {{- with .Values.config.values.infra }}
-    {{- range $key, $value := . }}
-    {{ $key }}: {{ $value | quote }}
```
ewkoch, line 24: This broke for some reason, I think related to how helm / go templating was handling certain types. I also don't like it because it doesn't clarify what actually needs to be in the config

charts/model-engine/templates/service_config_map.yaml
```diff
@@ -11,46 +12,110 @@ metadata:
...
+          ]
+      }
+
+    sqs_queue_tag_template: |-
+      {
+          "infra.scale.com/product": "{{ .Values.productTag }}",
+          "infra.scale.com/team": "${team}",
+          "infra.scale.com/contact": "{{ .Values.contactEmail }}",
+          "infra.scale.com/customer": "AllCustomers",
+          "infra.scale.com/financialOwner": "{{ .Values.contactEmail}}",
```
ewkoch, line 71: need to template "infra.scale.com" to something or allow these as additional tags set in the values.yaml

charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -95,14 +95,17 @@ data:
...
           {{- end }}
           {{- if eq $device "gpu" }}
-          {{- if empty $node_selector }}
-          nodeSelector:
-          {{- end }}
-            k8s.amazonaws.com/accelerator: ${GPU_TYPE}
+          # {{- if empty $node_selector }}
+          # nodeSelector:
+          # {{- end }}
+          #   k8s.amazonaws.com/accelerator: ${GPU_TYPE}
```
ewkoch, line 101: will re-add this, just serves as a note that need to add documentation that this tag must be set on nodegroups

charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -95,14 +95,17 @@ data:
...
           tolerations:
             - key: "nvidia.com/gpu"
               operator: "Exists"
               effect: "NoSchedule"
+            - key: "gpu_a100_multi"
+              operator: "Exists"
+              effect: "NoSchedule"
```
ewkoch, line 108: hack, need to delete

charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -522,6 +525,7 @@ data:
         loadBalancer:
           simple: LEAST_REQUEST
   {{- end }}
+  {{- if and (.Capabilities.APIVersions.Has "autoscaling.k8s.io/v1") (.Values.autoscaling.vertical.enabled) }}
```
ewkoch, line 528: this is important if you don't have autoscaling

charts/model-engine/values.yaml
```diff
@@ -5,7 +5,7 @@ redis:
   auth:
 db:
   runDbInitScript: false
-balloonNodeSelector:
-  node-lifecycle: normal
-nodeSelector:
-  node-lifecycle: normal
+# balloonNodeSelector:
```
ewkoch, line 8: note- need to build out a good default values.yaml with comments specifying the purpose of each configuration option.

charts/model-engine/templates/service_account_inference.yaml
```diff
@@ -1,25 +0,0 @@
-{{- if and (.Values.serviceTemplate) (.Values.serviceTemplate.createServiceAccount) (.Values.serviceTemplate.serviceAccountAnnotations) (.Values.serviceTemplate.serviceAccountName) (.Values.config.values.launch.endpoint_namespace)}}
```
seanshi-scale, line 1: is there a reason we're removing this? seems like we could just add some more conditions to the if to achieve the same effect

charts/model-engine/templates/service_config_map.yaml
seanshi-scale: imo we shouldn't hardcode the values that go into launch_service_config and infra_service_config; doing so introduces a coupling between the helm chart and the model engine code, e.g. if the values that need to go in either of the two configs changes. Probably good to see what ended up breaking though and fixing that

charts/model-engine/templates/_helpers.tpl
dmchoiboi: Yes let's apply this change

charts/model-engine/templates/inference_framework_config.yaml
dmchoiboi: these are different images so we would want vllmTag, vllmBatchTag, vllmBatchV2Tag; and yes let's default to latest

charts/model-engine/templates/istio-virtualservice.yaml
dmchoiboi: curious to see what the rest of the dns refactor changes look like

charts/model-engine/templates/inference_framework_config.yaml
ewkoch: done

charts/model-engine/templates/istio-virtualservice.yaml
ewkoch: i made the rest of the refactor- it helps when model engine is a subchart to prevent duplication of changing the core domain a bunch of places

charts/model-engine/templates/service_config_map.yaml
ewkoch: i would prefer a single source of truth for configuring the deployment of a new instance- manually changing the yaml files is also a coupling between config and code, and similarly error prone-

i think there's probably a bigger refactor necessary either way to reduce the complexity of interacting with those configs

charts/model-engine/templates/inference_framework_config.yaml
dmchoiboi: ah, internally, we have a separate upgrade process for vllm that doesn't require install helm charts. If you're interested, we can socialize that a bit better

charts/model-engine/templates/service_account_inference.yaml
ewkoch: yea i'm fixing this- didn't realize this "createServiceAccount" was just for the inference account, which i was getting a naming conflict with so i removed for a quick fix

just updating with explicit naming, so we create another model-engine service account by default

charts/model-engine/values.yaml
```diff
@@ -1,11 +1,313 @@
...
+tolerations: [ ]
+affinity: { }
+
+# Specifies the configuration on the Gateway service kube service
+service:
+  type: ClusterIP
+  port: 80
+
+# Creates istio virtual services for the Model Engine using the global domain nome and gateway specified below
+virtualService:
```
dmchoiboi, line 59: i think this is a backwards incompatible change; can we leave this as virtualservice

charts/model-engine/values.yaml
```diff
@@ -1,11 +1,313 @@
-dd_trace_enabled: true
-spellbook:
-  enabled: false
-redis:
-  auth:
+# If specified, will override the name of the deployed services
```
dmchoiboi, line 1: Thanks for the comprehensive documentation here. It's very much appreciated.

charts/model-engine/templates/service_config_map.yaml
dmchoiboi: can pull your changes to service/infra configs to a separate file for now for documentation purpose as we figure out a proper way to do this

charts/model-engine/templates/service_config_map.yaml
```diff
@@ -101,8 +101,8 @@ data:
     env: {{ .Values.context | quote }}
     cloud_provider: "aws"
     env: "prod"
-    k8s_cluster_name: "usgw1-prod"
-    dns_host_domain: "model-engine.ml-serving.{{ $.Values.global.networking.internalDomain }}"
```
seanshi-scale: oh hmm this seems like this was a bug

charts/model-engine/templates/inference_framework_config.yaml
ewkoch: oh rip i'll wrap this in a config then, it's easiest for use to change with upgrades but curious about your process

charts/model-engine/templates/istio-virtualservice.yaml
```diff
@@ -1,24 +1,20 @@
-{{- if .Values.virtualservice.enabled -}}
+{{- if .values.virtualService.enabled -}}
```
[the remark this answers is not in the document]
ewkoch: no thanks for the catch

PR #653, opened 2024-10-30
Add script to stamp initial schema

dmchoiboi:
# Pull Request Summary

Needed for graceful helm upgrade for environments that were deployed before we introduced alembic

## Test Plan and Usage Guide


seanshi-scale (approved):
how does the script get run? i.e. do we have to manually run it or something like that?

dmchoiboi:
>  do we have to manually run it or something like that?
yes, it will need to be run manually before applying a new helm upgrade

PR #654, opened 2024-10-31
Remove ENV requirement for db migration

dmchoiboi:
# Pull Request Summary

This requirement is too strict. ENV is not a guaranteed environemntal variable

## Test Plan and Usage Guide


jasonyang101 (approved):

seanshi-scale (approved):
think we just need infra_config().env to exist, which I guess is always true

PR #660, opened 2024-11-13
emit model name in dd traces, also emit error dd metrics on http timeouts

seanshi-scale:
# Pull Request Summary

See title.
Noticed that we don't have much granularity on errors happening if sync/streaming endpoints don't return, so am emitting that

I am calling a gateway from another gateway which is maybe not the best, but the alternative was to stick it in the use case and parse out the stream returned from the StreamingInferenceGateway, which seemed worse.

## Test Plan and Usage Guide

deploy to cluster, send requests to an endpoint that times out, see if we get the metrics appearing, which we do

### DIFF
--- a/model-engine/model_engine_server/domain/gateways/streaming_model_endpoint_inference_gateway.py
+++ b/model-engine/model_engine_server/domain/gateways/streaming_model_endpoint_inference_gateway.py
@@ -21,7 +21,7 @@ class StreamingModelEndpointInferenceGateway(ABC):
         topic: str,
         predict_request: SyncEndpointPredictV1Request,
         manually_resolve_dns: bool,
-        endpoint_name: Optional[str],
+        endpoint_name: Optional[str] = None,
     ) -> AsyncIterable[SyncEndpointPredictV1Response]:
         """
         Runs a prediction request and returns a streaming response.
--- a/model-engine/model_engine_server/domain/gateways/streaming_model_endpoint_inference_gateway.py
+++ b/model-engine/model_engine_server/domain/gateways/streaming_model_endpoint_inference_gateway.py
@@ -21,7 +21,7 @@ class StreamingModelEndpointInferenceGateway(ABC):
         topic: str,
         predict_request: SyncEndpointPredictV1Request,
         manually_resolve_dns: bool,
-        readable_endpoint_name: Optional[str],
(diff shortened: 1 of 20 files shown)


model-engine/model_engine_server/infra/gateways/datadog_monitoring_metrics_gateway.py
```diff
@@ -85,3 +85,8 @@ def emit_token_count_metrics(self, token_usage: TokenUsage, metadata: MetricMeta
         inter_token_latency = f"{self.prefix}.inter_token_latency"
         if token_usage.inter_token_latency is not None:
             statsd.distribution(inter_token_latency, token_usage.inter_token_latency, tags=tags)
+
+    def emit_http_call_error_metrics(self, endpoint_name: str, error_code: int):
+        tags = self.tags
+        tags.extend([f"endpoint_name:{endpoint_name}", f"error_code:{error_code}"])
+        statsd.increment(f"{self.prefix}.timeout", tags=tags)
```
seanshi-scale, line 92: TODO rename this guy

model-engine/model_engine_server/domain/gateways/streaming_model_endpoint_inference_gateway.py
```diff
@@ -17,7 +17,11 @@ class StreamingModelEndpointInferenceGateway(ABC):
 
     @abstractmethod
     def streaming_predict(
-        self, topic: str, predict_request: SyncEndpointPredictV1Request, manually_resolve_dns: bool
+        self,
+        topic: str,
+        predict_request: SyncEndpointPredictV1Request,
+        manually_resolve_dns: bool,
+        readable_endpoint_name: Optional[str],
```
dmchoiboi, line 24: where does this get used?

dmchoiboi: oh sorry this is an abstract class

dmchoiboi, line 24: can we just use `endpoint_name` instead of `readable_endpoint_name`?

dmchoiboi (approved):
small nit about naming

PR #661, opened 2024-11-18
Add max_model_len as Optional Argument for Model.create API

sghanta98:
Pass said parameter to vLLM engine if requested by user

### DIFF
--- a/clients/python/llmengine/__init__.py
+++ b/clients/python/llmengine/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.0.0beta43"
+__version__ = "0.0.0beta44"
 
 import os
 from typing import Sequence
--- a/clients/python/llmengine/model.py
+++ b/clients/python/llmengine/model.py
@@ -42,6 +42,7 @@ class Model(APIEngine):
         num_shards: int = 1,
         quantize: Optional[Quantization] = None,
         checkpoint_path: Optional[str] = None,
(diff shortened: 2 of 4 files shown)


yixu34:
Any tests to update/create?

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -290,6 +290,18 @@
     LLMInferenceFramework.TENSORRT_LLM: [],
 }
 
+# This was removed in ba68b8da2f6ef1f6df3f703b35456b72fbe20176. Do we need this if we are not
+# validating max_model_len?
+_MAX_MODEL_LENGTH_CONFIG: Dict[str, Dict[str, Optional[int]]] = {
```
dmchoiboi, line 295: let's remove; we don't want to be maintaining a hardcoded list of configs since things change so much. Given that our defaults work out of box, we can assume that our users are educated enough to provide the right model length if they need to deviate from default behavior

sghanta98: Fair enough 👍

sghanta98:
> Any tests to update/create?

There seems to be an already existing test `test_create_model_endpoint_w_vllm_args` which tests with max-model-len parameter value. @dmchoiboi do you think a new test should be created?

dmchoiboi (approved):

dmchoiboi (approved):

PR #664, opened 2024-12-05
Don't fail async task on pod termination

seanshi-scale:
# Pull Request Summary

Kick requests back to the queue if an async worker is shutting down, by seeing if we hit a ConnectionError.
Note: If a request gets kicked back to the queue, the status goes from STARTED to PENDING. This masks a "RETRY" status from Celery. I didn't want to break backwards compatibility.

## Test Plan and Usage Guide

Put a test deployment on the cluster and made requests to it while scaling down the workers. The tasks eventually all succeed. Before, some tasks would fail.
image

### DIFF
--- a/model-engine/.coveragerc
+++ b/model-engine/.coveragerc
@@ -1,0 +1,2 @@
+[report]
+omit = model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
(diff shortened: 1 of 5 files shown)


model-engine/.coveragerc
```diff
@@ -0,0 +1,2 @@
+[report]
+omit = model_engine_server/infra/gateways/celery_task_queue_gateway.py
```
seanshi-scale, line 2: Can't really test this file since it's so intertwined with celery

dmchoiboi (approved):

PR #665, opened 2024-12-09
Make cooldown period configurable

sohumh:
# Pull Request Summary

Allow the `cooldownPeriod` used in keda for autoscaling to be set as a helm value.

## Test Plan and Usage Guide

Ran `helm template model-engine model-engine/ -f model-engine/values_circleci.yaml` to validate it gets populated as intended

### DIFF
--- a/charts/model-engine/values.yaml
+++ b/charts/model-engine/values.yaml
@@ -16,3 +16,5 @@ balloonNodeSelector:
   node-lifecycle: normal
 nodeSelector:
   node-lifecycle: normal
+keda:
+  cooldownPeriod: 300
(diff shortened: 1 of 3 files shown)


yunfeng-scale (approved):

charts/model-engine/values.yaml
```diff
@@ -16,3 +16,5 @@ balloonNodeSelector:
   node-lifecycle: normal
 nodeSelector:
   node-lifecycle: normal
+keda:
```
yunfeng-scale, line 19: can you also add it to value_samples and everywhere else, otherwise please set default values

sohumh: is there a list of all files for where the update should be made? I based my logic of this [PR](https://github.com/scaleapi/llm-engine/pull/651) which seems to only update one values file

yunfeng-scale: actually i think it might be fine. helm should be able to take default values from values.yaml

sohumh: ok great, will merge then. appreciate the review!

PR #666, opened 2024-12-10
API to restart deployment

yunfeng-scale:
# Pull Request Summary

as title. will return error for endpoints with more than 1 node

## Test Plan and Usage Guide

unit test and tested with prod deployments

### DIFF
--- a/model-engine/model_engine_server/api/model_endpoints_v1.py
+++ b/model-engine/model_engine_server/api/model_endpoints_v1.py
@@ -20,6 +20,7 @@ from model_engine_server.common.dtos.model_endpoints import (
     GetModelEndpointV1Response,
     ListModelEndpointsV1Response,
     ModelEndpointOrderBy,
+    RestartModelEndpointV1Response,
     UpdateModelEndpointV1Request,
     UpdateModelEndpointV1Response,
 )
@@ -42,6 +43,7 @@ from model_engine_server.domain.use_cases.model_endpoint_use_cases import (
     DeleteModelEndpointByIdV1UseCase,
     GetModelEndpointByIdV1UseCase,
     ListModelEndpointsV1UseCase,
+    RestartModelEndpointV1UseCase,
     UpdateModelEndpointByIdV1UseCase,
 )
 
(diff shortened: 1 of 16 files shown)


model-engine/model_engine_server/infra/services/live_model_endpoint_service.py
```diff
@@ -403,5 +403,26 @@ async def delete_model_endpoint(self, model_endpoint_id: str) -> None:
...
         logger.info(f"Endpoint delete released lock for {created_by}, {name}")
 
+    async def restart_model_endpoint(self, model_endpoint_id: str) -> None:
+        record = await self.model_endpoint_record_repository.get_model_endpoint_record(
+            model_endpoint_id=model_endpoint_id
+        )
+        if record is None:
+            raise ObjectNotFoundException
+
+        async with self.model_endpoint_record_repository.get_lock_context(record) as lock:
```
seanshi-scale, line 413: not super sure if we need the lock? the lock's been troublesome before

seanshi-scale: and if we can avoid it then it'd be nice

model-engine/model_engine_server/infra/gateways/resources/k8s_endpoint_resource_delegate.py
```diff
@@ -1489,6 +1494,33 @@ async def _delete_keda_scaled_object(endpoint_id: str) -> bool:
...
+                    "metadata": {
+                        "annotations": {
+                            "model-engine.scale.com/restartedAt": datetime.datetime.now().isoformat()
+                        }
+                    }
+                }
+            }
+        }
+
+        try:
```
seanshi-scale, line 1513: should we be more explicit in the LeaderWorkerSet case? e.g. return that "it doesn't work for LWSs"

yunfeng-scale: make sense. let me do this in the use case code

model-engine/model_engine_server/api/model_endpoints_v1.py
```diff
@@ -207,3 +209,27 @@ async def delete_model_endpoint(
...
+    """
+    Restarts the Model endpoint.
+    """
+    logger.info(f"POST /model-endpoints/{model_endpoint_id}/restart for {auth}")
+    try:
+        use_case = RestartModelEndpointV1UseCase(
+            model_endpoint_service=external_interfaces.model_endpoint_service,
+        )
+        return await use_case.execute(user=auth, model_endpoint_id=model_endpoint_id)
+    except (ObjectNotFoundException, ObjectNotAuthorizedException) as exc:
```
seanshi-scale, line 231: we probably want to catch the ObjectHasInvalidValueException here and return 400 (or something)?

seanshi-scale (approved):
had a comment about catching the InvalidValueException in the API and forwarding the response code (I don't think we have anything else that catches it?), other than that LGTM

PR #669, opened 2024-12-11
Make async workers able to work on more than 1 concurrent request

seanshi-scale:
# Pull Request Summary

add concurrency option to async forwarder and pass this through our apis

Follow up work:
figure out how well eventlet/gevent would work with the celery forwarder to increase our concurrency
add this option to sync

note: seems like we have about a limit of 20 concurrency with the prefork set up, so we might have to limit it there

## Test Plan and Usage Guide

unit tests

deploy on cluster:
- on create, it defaults to per_worker if that's low enough, and MAX if too high
- on create and update, it being too high fails validation (use case create) (done)
- on update, not changing concurrency makes it stay the same
- on update, changing it gets sent through
- the relevant worker can handle more than 1 concurrent requests, controlled by this concurrency value


clients/python/llmengine/data_types/rest.py
```diff
@@ -25,6 +25,7 @@ class ModelEndpointDeploymentState(BaseModel):
     min_workers: int = Field(..., ge=0)
     max_workers: int = Field(..., ge=0)
     per_worker: int = Field(..., gt=0)
+    concurrent_requests: int = Field(..., gt=0)
```
seanshi-scale, line 28: todo this needs to be thread through some api dtos

seanshi-scale: should be done

charts/model-engine/templates/service_template_config_map.yaml
```diff
@@ -233,7 +233,7 @@ data:
                 - "${SQS_QUEUE_URL}"
                 {{- end }}
                 - --num-workers
-                - "${PER_WORKER}"
+                - "${CONCURRENT_REQUESTS}"
```
yunfeng-scale, line 236: i think people have been pretty confused about "per_worker", possible to add documentation as much as possible to distinguish the two fields? concurrent_requests_per_worker?

seanshi-scale: yup, will add the documentation, probably to some other site

dmchoiboi (approved):

PR #671, opened 2024-12-26
Multinode batch inference

seanshi-scale:
# Pull Request Summary

Add the capability to do multinode inference

## Test Plan and Usage Guide

Tested the image on a cluster with 2xp5.48xlarges, was able to get results back.
Also works e2e


model-engine/model_engine_server/inference/vllm/init_ray_batch_inf_v2.py
```diff
@@ -0,0 +1,215 @@
+import argparse
```
seanshi-scale, line 1: lmk if this file is confusing

model-engine/model_engine_server/inference/vllm/vllm_batch.py
```diff
@@ -327,6 +339,40 @@ async def handle_batch_job(request: CreateBatchCompletionsEngineRequest) -> None
             trust_remote_code=request.model_cfg.trust_remote_code or False,
         )
 
+    if multinode:
+        job_completion_index = int(os.environ.get("JOB_COMPLETION_INDEX", 0))
+        # Initialize the ray cluster
+        leader_addr = os.environ.get("LEADER_ADDR")
+        leader_port = os.environ.get("LEADER_PORT")
```
yunfeng-scale, line 346: where do these coming from? are they from train?

model-engine/model_engine_server/inference/vllm/init_ray_batch_inf_v2.py
```diff
@@ -0,0 +1,221 @@
...
+    print(f"Timeout waiting for cluster to reach size {expected_nodes}", flush=True)
+    return False
+
+
+async def wait_for_head_node_to_exit() -> None:
+    # Runs in worker node
+    # Waits until there is no longer a connection to the head Ray node, then exits
+    # This name needs to equal the name of the file!
+    # Also, if this file gets moved to a different directory in the docker image, the path needs to be updated
+    worker_process = subprocess.Popen(
```
yunfeng-scale, line 88: why does it needs to be a script run, instead of calling `wait_for_head_node_to_exit_process` directly?

model-engine/model_engine_server/inference/vllm/vllm_batch.py
```diff
@@ -247,6 +254,9 @@ async def init_engine(
     default_engine_args_dict = dict(
         model=model,
         tensor_parallel_size=request.model_cfg.num_shards,
+        pipeline_parallel_size=int(
+            os.environ.get("NUM_INSTANCES", 1)
+        ),  # TODO maybe do something other than TP=8, PP=number of nodes
```
yunfeng-scale, line 259: yeah i wonder if TP/PP could be specified from request?

yunfeng-scale (approved):

model-engine/model_engine_server/inference/vllm/vllm_batch.py
seanshi-scale: yup, they're from train

model-engine/model_engine_server/inference/vllm/init_ray_batch_inf_v2.py
seanshi-scale: `wait_for_head_node_to_exit_process` returns a non-zero status code, which gets interpreted as an error in Train, which isn't the best ux, so I did this thing so that it would return successfully

yunfeng-scale: i see. can you put this in comment?

model-engine/model_engine_server/inference/vllm/vllm_batch.py
seanshi-scale: think we'd want to probably make sure if something other than tp=8 pp=2 works, I tried tp=16 at some point (I think with some vllm 0.5.x version? not sure) and it didn't work; once that's there then yeah I think it would make sense to do this

yunfeng-scale: i think multinode only makes sense with tp=16 + EFA / IB

PR #672, opened 2025-01-02
[SGP] Add cooldown period customizability for celery through annotations

sohumh:
# Pull Request Summary

Use the `celery.scaleml.autoscaler/cooldownPeriodSeconds` annotation to customize the cooldown period (how long the deployment stays up for after it has been autoscaled up).

For the initial version, will just rely on editing the deployment to add this annotation.

## Test Plan and Usage Guide

### DIFF
--- a/charts/model-engine/Chart.yaml
+++ b/charts/model-engine/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.6
+version: 0.1.7
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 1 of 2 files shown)


charts/model-engine/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.6
+version: 0.1.7
```
seanshi-scale, line 18: we shouldn't need to increment the chart version for the changes in this pr specifically, were there changes to the chart that didn't get the version increment?

sohumh: ah good catch, nope will remove

seanshi-scale (approved):

PR #676, opened 2025-01-14
Make the async task and sync task api actually return HTTP response codes from the user container

seanshi-scale:
# Pull Request Summary

Add a "status_code" field to the async and sync task responses, that gets filled in based on the value of whatever the user inference container returns. Note that streaming responses also have the status code returned. LLM streaming also still works

~As a corollary, the async task response turns from {task_id, status, traceback, result:{result:"serialized json"}} to {task_id, status, traceback, result:{result:"serialized json", status_code: int}}. I've decided not to filter it out since I doubt that it'll break things. TBD though~

upd: I've decided to filter it out, this does introduce somewhat of a footgun in the sense that the async forwarder needs to have wrap_response turned on, otherwise we may accidentally delete a `status_code` field inside of some user endpoint. At this point I doubt we'll change wrap_response and turn it off, and IMO a better way to do things is to manually unwrap inside the gateway if we ever decide to migrate the format to have less wrapping

## Test Plan and Usage Guide

Set up async/sync/stream endpoints using old code, and using new code. Deploy test gateway on cluster. Make requests to async/sync/stream and make sure that the status code appears in the task response for new code, and that something reasonable is returned for old code.

Also unit tests


integration_tests/rest_api_utils.py
```diff
@@ -968,12 +968,19 @@ def ensure_nonzero_available_llm_workers(endpoint_name: str, user_id: str):
...
+):
     print(response)
     assert response["status"] == "SUCCESS"
     assert response["traceback"] is None
     if return_pickled:
         assert response["result"]["result_url"].startswith("s3://")
+    elif expect_status_code_in_response:
+        # Limitation of async tasks is we need to stick the status code in the result
+        # and we don't yet manually filter it out
+        # Other option is to filter it out explicitly in the gateway
```
seanshi-scale, line 982: tbd whether I filter it out or not

integration_tests/test_endpoints.py
```diff
@@ -217,7 +217,7 @@ def test_sync_streaming_model_endpoint(capsys):
                 )
             )
             for response in task_responses:
-                ensure_inference_task_response_is_correct(response, False)
+                ensure_inference_task_response_is_correct(response, False, True)
```
seanshi-scale, line 220: todo False

dmchoiboi (approved):

PR #681, opened 2025-01-27
SGLang support

dmchoiboi:
# Pull Request Summary

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
@@ -472,7 +472,7 @@ class CreateLLMModelBundleV1UseCase:
                         chat_template_override,
                         additional_args=additional_vllm_args,
                     )
-            case LLMInferenceFramework.SGLANG:
+            case LLMInferenceFramework.SGLANG:  # pragma: no cover
                 if not hmi_config.sglang_repository:
                     raise ObjectHasInvalidValueException("SGLang repository is not set.")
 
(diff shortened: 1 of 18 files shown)


model-engine/model_engine_server/inference/service_requests.py
```diff
@@ -95,7 +95,9 @@ def _make_async_request(queue: str, args: List[Any], kwargs: Dict[str, Any]):
         args=[dict(cloudpickle=request_body), True],
         queue=queue,
     )
-    with allow_join_result():  # TODO this can cause deadlocks. Is the a way to identify it on this level?
+    with (
```
seanshi-scale, line 98: lmao this feels like extremely dead code

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -884,6 +777,108 @@ async def create_deepspeed_bundle(
...
+        #     final_weights_folder,
+        #     trust_remote_code=sglang_args.trust_remote_code or False,
+        # )
+
+        sglang_args.tp_size = num_shards
+
+        if chat_template_override:
+            sglang_args.chat_template = chat_template_override
+
+        sglang_cmd = f"python3 -m sglang.launch_server --model-path {huggingface_repo} --served-model-name {model_name} --port 5005 --host '::'"
```
seanshi-scale, line 823: does this mean that (for now) we have to download weights from HF each time a worker starts up? (i.e. are there plans to put stuff in s3 later on)

seanshi-scale (approved):
lgtm, think having sglang not download from s3 is fine for now

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
dmchoiboi: yes.. short term unblocking. I was using the official sglang image and obv s5cmd is not available there. We would need to do something similar to what we're doing w/ vllm

PR #683, opened 2025-02-03
Google Cloud Storage Integration

anishxyz:
# Pull Request Summary

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/infra/gateways/gcs_file_storage_gateway.py
+++ b/model-engine/model_engine_server/infra/gateways/gcs_file_storage_gateway.py
@@ -1,0 +1,104 @@
+import os
+from typing import List, Optional
+
+from google.cloud import storage
+
+from model_engine_server.core.config import infra_config
+from model_engine_server.domain.gateways.file_storage_gateway import (
+    FileMetadata,
+    FileStorageGateway,
+)
+from model_engine_server.infra.gateways.gcs_filesystem_gateway import GCSFilesystemGateway
+
+
+def get_gcs_key(owner: str, file_id: str) -> str:
+    """
+    Constructs a GCS object key from the owner and file_id.
+    """
+    return os.path.join(owner, file_id)
+
+
+def get_gcs_url(owner: str, file_id: str) -> str:
+    """
+    Returns the gs:// URL for the bucket, using the GCS key.
+    """
+    return f"gs://{infra_config().gcs_bucket}/{get_gcs_key(owner, file_id)}"
+
+
+class GCSFileStorageGateway(FileStorageGateway):
+    """
+    Concrete implementation of a file storage gateway backed by GCS.
+    """
+
+    def __init__(self):
+        self.filesystem_gateway = GCSFilesystemGateway()
+
+    async def get_url_from_id(self, owner: str, file_id: str) -> Optional[str]:
+        """
+        Returns a signed GCS URL for the given file.
+        """
+        try:
+            return self.filesystem_gateway.generate_signed_url(get_gcs_url(owner, file_id))
+        except Exception:
+            return None
+
+    async def get_file(self, owner: str, file_id: str) -> Optional[FileMetadata]:
+        """
+        Retrieves file metadata if it exists. Returns None if the file is missing.
+        """
+        try:
+            client = self.filesystem_gateway.get_storage_client({})
+            bucket = client.bucket(infra_config().gcs_bucket)
+            blob = bucket.blob(get_gcs_key(owner, file_id))
+            blob.reload()  # Fetch metadata
+            return FileMetadata(
+                id=file_id,
+                filename=file_id,
+                size=blob.size,
+                owner=owner,
+                updated_at=blob.updated,
+            )
+        except Exception:
+            return None
+
+    async def get_file_content(self, owner: str, file_id: str) -> Optional[str]:
+        """
+        Reads and returns the string content of the file.
+        """
+        try:
+            with self.filesystem_gateway.open(get_gcs_url(owner, file_id)) as f:
+                return f.read()
+        except Exception:
+            return None
+
+    async def upload_file(self, owner: str, filename: str, content: bytes) -> str:
+        """
+        Uploads the file to the GCS bucket. Returns the filename used in bucket.
+        """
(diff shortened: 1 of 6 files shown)


model-engine/model_engine_server/infra/gateways/gcs_file_storage_gateway.py
```diff
@@ -0,0 +1,104 @@
...
+
+from model_engine_server.core.config import infra_config
+from model_engine_server.domain.gateways.file_storage_gateway import (
+    FileMetadata,
+    FileStorageGateway,
+)
+from model_engine_server.infra.gateways.gcs_filesystem_gateway import GCSFilesystemGateway
+
+
+def get_gcs_key(owner: str, file_id: str) -> str:
```
yixu34, line 14: nit: I'd prefix these w/ an underscore so that no one is tempted to try and import these from outside this file, thus breaking Clean Architecture norms.

model-engine/model_engine_server/infra/gateways/gcs_file_storage_gateway.py
```diff
@@ -0,0 +1,104 @@
...
+        except Exception:
+            return None
+
+    async def get_file(self, owner: str, file_id: str) -> Optional[FileMetadata]:
+        """
+        Retrieves file metadata if it exists. Returns None if the file is missing.
+        """
+        try:
+            client = self.filesystem_gateway.get_storage_client({})
+            bucket = client.bucket(infra_config().gcs_bucket)
```
yixu34, line 51: I know this pattern was already there, but I think it'd probably make more sense to pass in the bucket into the constructor of this class. This way, there's one less dependency on the old `infra_config` object. @seanshi-scale @tiffzhao5 thoughts?

Could also make the argument to just pass in the bucket as an argument with every `get_file` call, but that's outside of the scope of this change I'd say.

model-engine/model_engine_server/infra/gateways/gcs_llm_artifact_gateway.py
```diff
@@ -0,0 +1,115 @@
...
+        gcs = self._get_gcs_client(kwargs)
+        parsed_remote = parse_attachment_url(path, clean_key=False)
+        bucket_name = parsed_remote.bucket
+        prefix = parsed_remote.key
+
+        bucket = gcs.bucket(bucket_name)
+        blobs = bucket.list_blobs(prefix=prefix)
+        downloaded_files = []
+
+        for blob in blobs:
```
yixu34, line 54: Looks like this is just sequentially downloading the files? Is this the recommended way? ChatGPT showed me two responses, (1) with a ThreadPoolExecutor, and (2) this one.

model-engine/model_engine_server/infra/gateways/gcs_file_storage_gateway.py
seanshi-scale: I'm fine with having the bucket be passed in the constructor (in addition to anything else from any configs); `dependencies.py` does read in from infra_config at times to figure out constructor arguments, so there's precedent already

model-engine/model_engine_server/infra/gateways/gcs_filesystem_gateway.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    Concrete implementation for interacting with Google Cloud Storage.
+    """
+
+    def get_storage_client(self, kwargs: Optional[Dict]) -> storage.Client:
+        """
+        Retrieve or create a Google Cloud Storage client. Could optionally
+        utilize environment variables or passed-in credentials.
+        """
+        project = kwargs.get("gcp_project", os.getenv("GCP_PROJECT"))
```
seanshi-scale, line 20: where does this env var get set? it seems analogous to `AWS_PROFILE` but those changes would need to be baked into any relevant k8s yamls most likely

model-engine/model_engine_server/infra/gateways/gcs_llm_artifact_gateway.py
```diff
@@ -0,0 +1,115 @@
...
+        gcs = self._get_gcs_client(kwargs)
+        parsed_remote = parse_attachment_url(path, clean_key=False)
+        bucket_name = parsed_remote.bucket
+        prefix = parsed_remote.key
+
+        bucket = gcs.bucket(bucket_name)
+        files = [blob.name for blob in bucket.list_blobs(prefix=prefix)]
+        return files
+
+    def download_files(self, path: str, target_path: str, overwrite=False, **kwargs) -> List[str]:
```
seanshi-scale, line 41: IIUC with the current state of the code, this only gets called if you use TGI, LightLLM, TensorRT-LLM, Deepspeed as inference frameworks, so I doubt this code ends up getting exercised in practice (it's only used to download a tokenizer to count tokens on the Gateway)

model-engine/model_engine_server/infra/gateways/gcs_file_storage_gateway.py
seanshi-scale: btw I think the s3_file_storage_gateway also doesn't have the prefixes

model-engine/model_engine_server/infra/gateways/gcs_llm_artifact_gateway.py
```diff
@@ -0,0 +1,115 @@
...
+        bucket_name = parsed_remote.bucket
+        prefix = parsed_remote.key
+
+        bucket = gcs.bucket(bucket_name)
+        blobs = bucket.list_blobs(prefix=prefix)
+        downloaded_files = []
+
+        for blob in blobs:
+            # Remove prefix and leading slash to derive local name
+            file_path_suffix = blob.name.replace(prefix, "").lstrip("/")
```
seanshi-scale, line 56: do you want to `replace(prefix, "", count=1)`? (or something like that) just in case the prefix appears elsewhere in the string

for that matter is that also a bug in the s3 implementation?

seanshi-scale (commented):
looks good, just a few things that looked like bugs (from the s3 implementation)

anishxyz:
Hi! Sorry should have made this a draft pr but still need to test. I put it up mainly to coordinate with aagam

PR #687, opened 2025-02-27
Handle large batches w/ guided decoding

dmchoiboi:
# Pull Request Summary

Limit concurrency of sendings requests to the vLLM engine when requests have guided decoding since there is a per request initialization cost that results in request loading taking longer than the default `VLLM_ENGINE_ITERATION_TIMEOUT_S` of 30s.

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/inference/vllm/vllm_batch.py
+++ b/model-engine/model_engine_server/inference/vllm/vllm_batch.py
@@ -58,7 +58,7 @@ from vllm.utils import merge_async_iterators
 
 CONFIG_FILE = os.getenv("CONFIG_FILE")
 AWS_REGION = os.getenv("AWS_REGION", "us-west-2")
-MODEL_WEIGHTS_FOLDER = os.getenv("MODEL_WEIGHTS_FOLDER", "./model_weights")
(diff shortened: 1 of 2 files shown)


model-engine/model_engine_server/inference/vllm/vllm_batch.py
```diff
@@ -191,6 +191,26 @@ async def generate_v1_completions(
     return outputs
 
 
+# This is needed to handle the cases where it takes too long to process all of the requests before
+# the configured 'VLLM_ENGINE_ITERATION_TIMEOUT_S' 30s timeout.
+def determine_max_concurrent_requests(
+    requests: Union[List[CompletionRequest], List[ChatCompletionRequest]]
+) -> int:
+    # Guided decoding
+    # For example, with guided decoding, vLLM initializes a guided decoding logit processor per request, and
```
toekneechinscale, line 200: Nice! It seems like multiple ppl on vllm had this issue with the logit processor and keep trying to change it. I think this workaround is pretty clean

toekneechinscale (approved):
LGTM thanks for looking into this! This will make our T&E offering so much better!

tiffzhao5 (approved):
nice!

PR #698, opened 2025-06-25
Add SGP tracing to model engine

neumark-scale:
# Adding SGP trace functionality to model engine

SGP provides an API for submitting traces, which can be used to log cross-server requests. When model-engine receives an inference request with an `x-sgp-trace-config` HTTP header, it can do one of two things:
1. Submit trace spans of it's own, logging events which occur within the model-engine API. These traces are collected by an SGP API.
2. Forward the `x-sgp-trace-config` to the model endpoint. For sync endpoints, this happens directly, for async endpoints, this value is included in the celery task queue message.


dmchoiboi (commented):
general comment: should remove SGP specific logic into a plugin and add generic capabilities to enable tracing

PR #699, opened 2025-07-14
add generic tracing support to model-engine

neumark-scale:
# Pull Request Summary

## Test Plan and Usage Guide

### DIFF
--- a/model-engine/model_engine_server/api/dependencies.py
+++ b/model-engine/model_engine_server/api/dependencies.py
@@ -1,4 +1,5 @@
 import asyncio
+from functools import lru_cache
 import os
 import time
 from dataclasses import dataclass
@@ -15,6 +16,7 @@ from model_engine_server.core.auth.fake_authentication_repository import (
     FakeAuthenticationRepository,
 )
 from model_engine_server.core.config import infra_config
+from model_engine_server.core.tracing.fake_tracing_gateway import FakeTracingGateway
 from model_engine_server.core.loggers import (
     LoggerTagKey,
     LoggerTagManager,
@@ -31,6 +33,7 @@ from model_engine_server.domain.gateways import (
     MonitoringMetricsGateway,
     TaskQueueGateway,
 )
+from model_engine_server.core.tracing.tracing_gateway import TracingGateway
 from model_engine_server.domain.repositories import (
     DockerImageBatchJobBundleRepository,
     DockerRepository,
@@ -166,6 +169,7 @@ class ExternalInterfaces:
     monitoring_metrics_gateway: MonitoringMetricsGateway
     tokenizer_repository: TokenizerRepository
     streaming_storage_gateway: StreamingStorageGateway
+    tracing_gateway: TracingGateway
 
 
 def get_default_monitoring_metrics_gateway() -> MonitoringMetricsGateway:
@@ -352,6 +356,8 @@ def _get_external_interfaces(
 
     streaming_storage_gateway = FirehoseStreamingStorageGateway()
 
+    tracing_gateway = FakeTracingGateway()
+
     external_interfaces = ExternalInterfaces(
         docker_repository=docker_repository,
         model_bundle_repository=model_bundle_repository,
@@ -376,6 +382,7 @@ def _get_external_interfaces(
         monitoring_metrics_gateway=monitoring_metrics_gateway,
         tokenizer_repository=tokenizer_repository,
         streaming_storage_gateway=streaming_storage_gateway,
+        tracing_gateway=tracing_gateway
     )
     return external_interfaces
 
@@ -392,6 +399,7 @@ def get_default_external_interfaces_read_only() -> ExternalInterfaces:
     return _get_external_interfaces(read_only=True, session=session)
 
 
+@lru_cache(maxsize=1)
 async def get_external_interfaces():
     try:
         from plugins.dependencies import get_external_interfaces as get_custom_external_interfaces
@@ -402,7 +410,7 @@ async def get_external_interfaces():
     finally:
         pass
 
-
+@lru_cache(maxsize=1)
 async def get_external_interfaces_read_only():
     try:
         from plugins.dependencies import (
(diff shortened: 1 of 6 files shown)


model-engine/model_engine_server/api/dependencies.py
```diff
@@ -352,6 +355,8 @@ def _get_external_interfaces(
 
     streaming_storage_gateway = FirehoseStreamingStorageGateway()
 
+    tracing_gateway = FakeTracingGateway()
```
dmchoiboi: we should not be using fakes in production. Can you either implement a default tracing gateway (e.g. LiveTracingGateway), or make this optional?

model-engine/model_engine_server/core/tracing/__init__.py
```diff
@@ -0,0 +1,18 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from model_engine_server.core.tracing.tracing_gateway import TracingGateway
+
+def get_tracing_gateway() -> "TracingGateway":
```
dmchoiboi, line 9: this is a bit of an antipattern; right now we rely on `get_external_interfaces` + dependency injection to ensure dependencies are instantiated once for the service. As is, this results in a tracing gateway initialized per retrieval.

model-engine/model_engine_server/core/tracing/tracing_gateway.py
```diff
@@ -0,0 +1,27 @@
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Any, Dict, Generator, Optional, Union
+from model_engine_server.core.tracing.span import Span
+
+if TYPE_CHECKING:
```
dmchoiboi, line 7: is there a reason we do this? As of now, we pretty much enforce typechecking and include all relevant libraries in our requirements

model-engine/model_engine_server/infra/gateways/live_sync_model_endpoint_inference_gateway.py
```diff
@@ -33,8 +33,10 @@
     stop_any,
     wait_exponential,
 )
+from model_engine_server.core.tracing import get_tracing_gateway
 
 logger = make_logger(logger_name())
+tracing_gateway = get_tracing_gateway()
```
dmchoiboi: instead of initializing here, pass as constructor arg in `LiveSyncModelEndpointInferenceGateway`

model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
```diff
@@ -38,6 +40,8 @@
 celery_servicebus = celery_app(
     None, broker_type=str(BrokerType.SERVICEBUS.value), backend_protocol=backend_protocol
 )
+tracing_gateway = get_tracing_gateway()
```
dmchoiboi: pass as arg into CeleryTaskQueueGateway

model-engine/model_engine_server/inference/forwarding/celery_forwarder.py
```diff
@@ -144,11 +146,16 @@ def exec_func(payload, arrival_timestamp, *ignored_args, **ignored_kwargs):
             # Don't fail the celery task even if there's a status code
             # (otherwise we can't really control what gets put in the result attribute)
             # in the task (https://docs.celeryq.dev/en/stable/reference/celery.result.html#celery.result.AsyncResult.status)
-            result = forwarder(payload)
-            request_duration = datetime.now() - arrival_timestamp
-            if request_duration > timedelta(seconds=DEFAULT_TASK_VISIBILITY_SECONDS):
-                monitoring_metrics_gateway.emit_async_task_stuck_metric(queue_name)
-            return result
+            if trace_config:
```
dmchoiboi: depending on whether tracing gateway is optional or required, should update this to only conditionally create the span?

model-engine/model_engine_server/inference/forwarding/celery_forwarder.py
```diff
@@ -24,9 +24,11 @@
     DatadogInferenceMonitoringMetricsGateway,
 )
 from requests import ConnectionError
+from model_engine_server.core.tracing import get_tracing_gateway
 
 logger = make_logger(logger_name())
 
+tracing_gateway = get_tracing_gateway()
```
dmchoiboi, line 31: hm if you want to use this paradigm in the forwarder (which doesn't have a dependency injection framework), we should update the function to manage a singleton object instead of re-initializing the gateway for this file and the `forwarding.py` file below

dmchoiboi (changes requested):
Thanks for making this change! Left some comments around proper gateway instantiation

model-engine/model_engine_server/core/tracing/__init__.py
ghost: The issue with dependency injenction is that the trace gateway is also used by code which run outside of fastapi (eg: celery forwarder). I added an `@lru_cache()` decorator which prevents the function from executing multiple times.

dmchoiboi: I think using this for things outside of fastapi is fine (like celery forwarder), but for the main app, lets use the dependency injection approach

model-engine/model_engine_server/inference/forwarding/celery_forwarder.py
ghost: The span can be created unconditionally, the default `LiveTracingGateway`'s span collection logic is a no-op, so creating the span is harmless.

model-engine/model_engine_server/inference/forwarding/celery_forwarder.py
ghost: I think the `@lru_cache` decorator should make this behave like a singleton.

model-engine/model_engine_server/core/tracing/tracing_gateway.py
ghost: The `if TYPE_CHECKING` clause makes the imports which follow conditional.
- If the code is running at type checking time (mypy / vscode LSP) then the block is executed
- Otherwise it's skipped. Since we're only importing types, this eliminates any errors which stem from failure to import the types.

dmchoiboi (approved):
lgtm

model-engine/model_engine_server/core/tracing/tracing_gateway.py
dmchoiboi, line 7: I think we can get rid of TYPE_CHECKING here as well?

PR #700, opened 2025-08-18
on prem changes to disable cloud solutions

jen-scymanski-scale:
# Pull Request Summary

Fixes Needed for model-engine to work on-prem
AWS Disable Logic: Added DISABLE_AWS=true environment variable support
Redis Fallback: S3 backend falls back to Redis when AWS is disabled
Lazy Initialization: Celery apps are initialized lazily to avoid import-time AWS sessions
On-Premises Redis: Added direct Redis connection support via ONPREM_REDIS_HOST
Configurable Gunicorn: Made worker timeouts and settings configurable via environment variables
Thread-Safe Logging: Fixed potential race conditions in logger initialization
Broker Disable Options: Added DISABLE_SQS_BROKER and DISABLE_SERVICEBUS_BROKER flags

## Test Plan and Usage Guide

Tested locally with scripts
AWS Disable Logic: Tested DISABLE_AWS=true prevents AWS session creation
Redis Fallback: Verified S3 backend falls back to Redis when AWS disabled
Lazy Initialization: Confirmed Celery apps don't create AWS sessions at import time
Thread-Safe Logging: Tested logger initialization in multi-threaded environments
Direct Redis Connection: Tested ONPREM_REDIS_HOST configuration
Server Startup: Confirmed server starts with all cloud services disabled
Environment Variables: Verified all new config variables work correctly
Async Task Workflow: Tested complete task submission and result retrieval
Broker Disable Logic: Verified SQS/ServiceBus fallback to Redis
Error Handling: Tested graceful handling of missing AWS credentials
Backward Compatibility: Confirmed existing cloud functionality still works
No Breaking Changes: Verified no existing APIs are broken

### DIFF
--- a/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
+++ b/model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
@@ -799,9 +799,10 @@ class CreateLLMModelBundleV1UseCase:
                                 "MODEL_NAME": model_name,
                             },
                             protocol="http",
-                            readiness_initial_delay_seconds=1800,
+                            readiness_initial_delay_seconds=120,
                         ),
                         metadata={},
+                        
                     ),
                     do_auth_check=False,
                 )
@@ -1049,7 +1050,7 @@ class CreateLLMModelBundleV1UseCase:
                 command=command,
                 streaming_command=command,
                 protocol="http",
-                readiness_initial_delay_seconds=1800,  # 30 minutes for large model downloads  
+                readiness_initial_delay_seconds=120,  # 30 minutes for large model downloads  
                 healthcheck_route="/health",  # Standard health check route
                 predict_route="/predict",
                 streaming_predict_route="/stream",
--- a/model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
+++ b/model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
@@ -1,6 +1,6 @@
 forwarder:
   sync:
-    user_port: 5005
+    user_port: e 5005
     user_hostname: "localhost"
     use_grpc: false
     predict_route: "/predict"
@@ -9,7 +9,7 @@ forwarder:
     model_engine_unwrap: true
     serialize_results_as_string: true
     forward_http_status: true
-    extra_routes: []
+    extra_routes: ["/v1/chat/completions", "/v1/completions"]
   stream:
     user_port: 5005
     user_hostname: "localhost"
@@ -18,5 +18,5 @@ forwarder:
     batch_route: null
     model_engine_unwrap: true
     serialize_results_as_string: false
-    extra_routes: []
+    extra_routes: ["/v1/chat/completions", "/v1/completions"]
   max_concurrency: 100
--- a/model-engine/model_engine_server/inference/vllm/vllm_adapter.py
+++ b/model-engine/model_engine_server/inference/vllm/vllm_adapter.py
@@ -1,6 +1,7 @@
 import argparse
 import json
 from typing import AsyncGenerator, Optional, List
+
 import uvicorn
 from fastapi import BackgroundTasks, FastAPI, Request
 from fastapi.responses import Response, StreamingResponse
@@ -9,35 +10,58 @@ from vllm.engine.async_llm_engine import AsyncLLMEngine
 from vllm.sampling_params import SamplingParams
 from vllm.utils import random_uuid
 from pydantic import BaseModel, Field
+from transformers import AutoTokenizer
+
+
 TIMEOUT_KEEP_ALIVE = 5  # seconds.
 TIMEOUT_TO_PREVENT_DEADLOCK = 1  # seconds
+tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-8B")
 app = FastAPI()
+
 class CompletionRequest(BaseModel):
     temperature: float = Field(default=0.2, ge=0.0, le=1.0)
     stop_sequences: Optional[List[str]] = Field(default=None, max_items=4)
     max_new_tokens: Optional[int] = Field(default=None)
     prompt: Optional[str] = Field(default=None)
     prompts: Optional[List[str]] = Field(default=None)
+
+class CompletionResponse(BaseModel):
+    completions: List[tuple]
+    prompt_tokens: int
+    response_tokens: int
+
 @app.get("/readyz")
 @app.get("/healthz") 
 @app.get("/health")
 def healthcheck():
     return "OK"
+
 @app.post("/predict")
 async def generate_egp_completions(request: Request) -> Response:
     request_id = random_uuid()
     request_dict = await request.json()
-    # Handle both Model-Engine format (prompt) and original format (prompts)
-    prompt = request_dict.pop("prompt", None)
-    if prompt:
-        prompts = [prompt]
-    else:
-        prompts = request_dict.pop("prompts", [])
+
+    chat_history = request_dict.pop("prompt", None)
+    if chat_history:
+        try:
+            messages = json.loads(chat_history)
+        except:
+            messages = [chat_history]
+    
     # Handle parameter name differences
     if "max_new_tokens" in request_dict:
         request_dict["max_tokens"] = request_dict.pop("max_new_tokens")
+
+    text = tokenizer.apply_chat_template(
+        messages,
+        tokenize=False,
+        add_generation_prompt=True,
+        enable_thinking=False,
+    )
+
     sampling_params = SamplingParams(**request_dict)
-    results_generator = engine.generate(prompts[0], sampling_params, request_id)
+    results_generator = engine.generate(text, sampling_params, request_id)
+
     final_output = None
     tokens = []
     last_output_text = ""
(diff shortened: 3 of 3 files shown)


model-engine/model_engine_server/common/config.py
```diff
@@ -104,6 +104,19 @@ def cache_redis_url(self) -> str:
             creds = get_key_file(self.cache_redis_aws_secret_name)  # Use default role
             return creds["cache-url"]
 
+        # Check if we're in an onprem environment with direct Redis access
+        if os.environ.get('ONPREM_REDIS_HOST'):
```
dmchoiboi: why do we not pass this in via config in the same way as the other redis configs?

jen-scymanski-scale: updated

charts/model-engine/values_onprem.yaml
```diff
@@ -0,0 +1,132 @@
+# values_onprem.yaml - On-premises deployment configuration
```
dmchoiboi: I don't think you need to include this file here? I believe SGP maintains their own values.yaml in their own repo somewhere cc @nicolastomeo ?

model-engine/model_engine_server/core/aws/secrets.py
```diff
@@ -14,17 +15,15 @@
...
-    try:
-        secret_value = json.loads(
-            secret_manager.get_secret_value(SecretId=secret_name)["SecretString"]
-        )
-        return secret_value
-    except ClientError as e:
-        logger.error(e)
-        logger.error(f"Failed to retrieve secret: {secret_name}")
-        return {}
+    response = secret_manager.get_secret_value(SecretId=secret_name)
```
dmchoiboi: I think we still want to do the try_catch wrapping to handle the cases where secret_manager client errors our

model-engine/model_engine_server/core/celery/app.py
```diff
@@ -530,18 +530,29 @@ def _get_backend_url_and_conf(
         # use db_num=1 for backend to differentiate from broker
         backend_url = get_redis_endpoint(1)
     elif backend_protocol == "s3":
-        backend_url = "s3://"
-        if aws_role is None:
-            aws_session = session(infra_config().profile_ml_worker)
+        # Check if AWS is disabled via config - if so, fall back to Redis backend
+        if infra_config().disable_aws:
```
dmchoiboi: instead of doing this, wondering if we address this upstream and figure out how to pass in "redis" as the backend_protocol in the on-prem scenario

charts/model-engine/templates/_helpers.tpl
```diff
@@ -369,6 +369,18 @@ env:
   - name: CIRCLECI
     value: "true"
   {{- end }}
+  {{- if .Values.gunicorn }}
```
dmchoiboi, line 372: anecdotally, we found it a lot easier to performance tune pure uvicorn, so we actually migrated most usage of gunicorn back to uvicorn. That being said, won't block your usage of it

model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
```diff
@@ -18,28 +19,61 @@
...
-celery_sqs = celery_app(
-    None,
-    s3_bucket=infra_config().s3_bucket,
-    broker_type=str(BrokerType.SQS.value),
-    backend_protocol=backend_protocol,
-)
-celery_servicebus = celery_app(
-    None, broker_type=str(BrokerType.SERVICEBUS.value), backend_protocol=backend_protocol
-)
+# Initialize celery apps lazily to avoid import-time AWS session creation
```
dmchoiboi, line 22: curious why we're not running into similar issues for our other non-AWS environments

model-engine/model_engine_server/entrypoints/start_fastapi_server.py
```diff
@@ -14,18 +15,25 @@ def start_gunicorn_server(port: int, num_workers: int, debug: bool) -> None:
     additional_args: List[str] = []
     if debug:
         additional_args.extend(["--reload", "--timeout", "0"])
+    
+    # Use environment variables for configuration with fallbacks
```
dmchoiboi, line 19: 👍

model-engine/model_engine_server/core/loggers.py
```diff
@@ -80,9 +80,24 @@ def make_standard_logger(name: str, log_level: int = logging.INFO) -> logging.Lo
         raise ValueError("Name must be a non-empty string.")
     logger = logging.getLogger(name)
     logger.setLevel(log_level)
-    logging.basicConfig(
-        format=LOG_FORMAT,
-    )
+    
+    # Thread-safe logging configuration - only configure if not already configured
```
dmchoiboi, line 84: interesting, was this manifesting in a particular error?

model-engine/model_engine_server/core/config.py
```diff
@@ -49,6 +49,17 @@ class _InfraConfig:
     firehose_role_arn: Optional[str] = None
     firehose_stream_name: Optional[str] = None
     prometheus_server_address: Optional[str] = None
+    # On-premises configuration
+    onprem_redis_host: Optional[str] = None
```
dmchoiboi: i'm wondering if there's an easy way to merge `onprem_redis_host` w/ the other `redis_host` arg that already exists.

model-engine/model_engine_server/core/config.py
```diff
@@ -49,6 +49,17 @@ class _InfraConfig:
...
+    onprem_redis_port: Optional[str] = "6379"
+    onprem_redis_password: Optional[str] = None
+    # AWS disable configuration
+    disable_aws: bool = False
```
dmchoiboi: could we instead add a new `cloud_provider == onprem` and change logic based off that?

model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
```diff
@@ -18,28 +19,61 @@
...
+            broker_type=str(BrokerType.SQS.value),
+            backend_protocol=backend_protocol,
+        )
+    return celery_sqs
+
+def _get_celery_servicebus():
+    global celery_servicebus
+    if celery_servicebus is None:
+        # Check if ServiceBus broker is disabled or if we're forcing Redis via config
+        if infra_config().disable_servicebus_broker or infra_config().force_celery_redis:
```
dmchoiboi: should we just throw an error if the app is configured to use sqs but the environment doesn't support it instead of adding a fallback logic?

model-engine/model_engine_server/core/loggers.py
jen-scymanski-scale: Yes, there was a specific recursive logging error causing worker crashes:

RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>

This occurred during Gunicorn worker startup when multiple processes tried to initialize logging simultaneously, causing thread-unsafe logging configuration and race conditions. The error led to worker crashes, which then triggered the WORKER TIMEOUT errors we were seeing.

The issue was that multiple Gunicorn workers starting at the same time would compete to write to stderr during logging setup, causing a reentrant call error that crashed the worker processes.

model-engine/model_engine_server/core/config.py
jen-scymanski-scale: We could consolidate this by using the existing `redis_host `field and adding logic to detect on-premises vs cloud environments. However, we kept them separate because:
`redis_host `is used for the message broker (Celery)
`onprem_redis_host `is used for the cache/result storage
They might point to different Redis instances in some deployments.. let me know if you would like me to combine them.

dmchoiboi: > redis_host is used for the message broker (Celery)
onprem_redis_host is used for the cache/result storage

This is a good distinction. I think it's better to make that more explicit in the naming as opposed to marking one as "onprem"

model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
jen-scymanski-scale: On Prem - This is likely due to no creds and the import failing
Container starts with NO AWS credentials
Python import hits celery_task_queue_gateway.py:19
boto3.Session() creation fails immediately
Import exception → container crash before application even starts

charts/model-engine/values_onprem.yaml
jen-scymanski-scale: ok, removed.

model-engine/model_engine_server/core/aws/roles.py
```diff
@@ -114,12 +115,17 @@ def assume_role(role_arn: str, role_session_name: Optional[str] = None) -> AwsCr
     )
 
 
-def session(role: Optional[str], session_type: SessionT = Session) -> SessionT:
+def session(role: Optional[str], session_type: SessionT = Session) -> Optional[SessionT]:
```
dmchoiboi: shouldn't need to touch this; this should only be used if cloud_provider == 'aws'

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -952,7 +984,7 @@ def _create_vllm_bundle_command(
             if hmi_config.sensitive_log_mode:
                 vllm_args.disable_log_requests = True
 
-            vllm_cmd = f"python -m vllm_server --model {final_weights_folder} --served-model-name {model_name} {final_weights_folder} --port 5005"
+            vllm_cmd = f"python -m vllm_server --model {final_weights_folder} --served-model-name {model_name} --port 5005"
```
sandeshghanta: @dmchoiboi - JFYI: this is where I've hackily changed it to support public vllm docker image.
```
            vllm_cmd = f"python3 -m vllm.entrypoints.openai.api_server --model {final_weights_folder} --served-model-name {model_name} --port 5005"
```

sandeshghanta: CC: @saeidbarati-scale

model-engine/model_engine_server/domain/use_cases/llm_model_endpoint_use_cases.py
```diff
@@ -1015,8 +1049,8 @@ async def create_vllm_bundle(
                 command=command,
                 streaming_command=command,
                 protocol="http",
-                readiness_initial_delay_seconds=10,
-                healthcheck_route="/health",
+                readiness_initial_delay_seconds=1800,  # 30 minutes for large model downloads
```
sandeshghanta: @saeidbarati-scale for context

PR #706, opened 2025-09-10
PSP-245 MCP/Passthrough Route Forwarder

MicahFulton:
# Pull Request Summary

Adding a forwarder for /mcp routes that allows passthrough.  This allows us to forward MCP requests to containers running running a service with MCP compatible requests.

A major difference between this and the other forwarders is it allows full request body and header passthrough to the downstream service, which is required for MCP servers which will look for specific headers.

https://linear.app/scale-epd/issue/PSP-245/mcp-model-engine-integration

## Test Plan and Usage Guide

Started up a test MCP Server locally with the command:
https://github.com/scaleapi/mcpx/tree/main/servers/weather
```
uv run weather.py
```

Started the forwarder locally with the command:

```
python model_engine_server/inference/forwarding/http_forwarder.py \
  --config model_engine_server/inference/configs/service--http_forwarder.yaml \
  --num-workers 1 \
  --host localhost \
  --port 8000
```

Added the weather tool to my cursor config:

```
"weather": {
      "url": "http://localhost:8000/mcp",
      "transport": "streamable-http"
    }
```

Validated that I could prompt: `can you use your weather tools to get my the weather for herndon va` and get a reponse.


model-engine/model_engine_server/inference/forwarding/forwarding.py
```diff
@@ -627,6 +628,92 @@ def endpoint(route: str) -> str:
         )
 
 
+@dataclass
+class MCPForwarder(ModelEngineSerializationMixin):
```
dmchoiboi: does the mcp server do anything different from the standard HttpForwarder?

MicahFulton: The main things are its a complete passthrough and doesn't modify headers, instead passes through and returns the headers to the underlying service, it also doesn't do any kind of serialization.

model-engine/model_engine_server/inference/forwarding/forwarding.py
```diff
@@ -627,6 +627,92 @@ def endpoint(route: str) -> str:
...
+        if self.user_hostname != "localhost":
+            raise NotImplementedError(
+                "Currently only localhost-based user-code services are supported with forwarders! "
+                f"Cannot handle {self.user_hostname=}"
+            )
+
+        def endpoint(route: str) -> str:
+            return f"http://{self.user_hostname}:{self.user_port}{route}"
+
+        passthrough_endpoint: str = endpoint("")
```
dmchoiboi: self.passthrough_route?

model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
```diff
@@ -19,4 +19,10 @@ forwarder:
     model_engine_unwrap: true
     serialize_results_as_string: false
     extra_routes: []
+  passthrough:
+    user_port: 5005
+    user_hostname: "localhost"
+    healthcheck_route: "/health"
+    passthrough_route: "/mcp"
```
dmchoiboi: i think we should either 

1. update passthrough_route to a list and remove extra_routes
2. just pass through all routes by default (excluding health)

model-engine/model_engine_server/inference/forwarding/forwarding.py
```diff
@@ -627,6 +627,92 @@ def endpoint(route: str) -> str:
...
+class PassthroughForwarder(ModelEngineSerializationMixin):
+    passthrough_endpoint: str
+
+    async def forward(self, request: Any):
+        async with aiohttp.ClientSession() as aioclient:
+            headers: dict[str, str] = dict(request.headers)
+            excluded_headers: set[str] = {
+                "host",
+                "content-length",
+                "transfer-encoding",
```
dmchoiboi: why do you get rid of transfer-encoding?

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -228,6 +265,11 @@ async def predict_or_stream(
     app.add_api_route(path="/readyz", endpoint=healthcheck, methods=["GET"])
     app.add_api_route(path="/predict", endpoint=predict, methods=["POST"])
     app.add_api_route(path="/stream", endpoint=stream, methods=["POST"])
+    app.add_api_route(
```
dmchoiboi: This should be loaded based off config file

model-engine/model_engine_server/inference/forwarding/forwarding.py
MicahFulton: removed it, since this is a passthrough the override shouldn't even exist.

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
MicahFulton: done

model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
MicahFulton: made it into a list

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -177,7 +218,7 @@ async def init_app():
     def healthcheck():
         return "OK"
 
-    def add_extra_routes(app: FastAPI):
+    def add_extra_stream_or_predict_routes(app: FastAPI):
```
dmchoiboi: it's sync_or_stream

model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
```diff
@@ -19,4 +19,9 @@ forwarder:
     model_engine_unwrap: true
     serialize_results_as_string: false
     extra_routes: []
+  passthrough:
```
dmchoiboi: remove? I think we're using the above config and providing forwarder_type now?

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -52,10 +57,23 @@ def get_streaming_forwarder_loader(
         del config["extra_routes"]
     if destination_path:
         config["predict_route"] = destination_path
+    if "forwarder_type" in config:
+        del config["forwarder_type"]
     streaming_forwarder_loader = LoadStreamingForwarder(**config)
     return streaming_forwarder_loader
 
 
+def get_stream_passthrough_forwarder_loader() -> LoadPassthroughForwarder:
```
dmchoiboi: will there be support for sync passthrough forwarder?

model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
MicahFulton: Its solely used as a template or list of properties that are available to pass through to the PassthroughForwarder.  We use this to filter out other config before passing to LoadPassthroughForwarder, but we could also just do that in the `get_streaming_forwarder_loader`

dmchoiboi: oh i didn't realize you read from this key when loading the passthrough forwarder

model-engine/model_engine_server/inference/forwarding/forwarding.py
dmchoiboi: oh i guess, is route="" intentional?

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -52,10 +57,23 @@ def get_streaming_forwarder_loader(
...
+    config = get_config().get("passthrough", {})
+    stream_config = get_config().get("stream", {})
```
dmchoiboi, line 70: I think i prefer that we merge the two configs

model-engine/model_engine_server/inference/forwarding/forwarding.py
MicahFulton: yea, since it should just map to the extra_routes passed through

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -224,6 +264,38 @@ async def predict_or_stream(
                 methods=["POST"],
             )
 
+    def add_stream_passthrough_routes(app: FastAPI):
+        config = get_config()
+
+        passthrough_forwarders: Dict[str, PassthroughForwarder] = dict()
+        for route in config.get("stream", {}).get("extra_routes", []):
+            passthrough_forwarders[route] = load_stream_passthrough_forwarder()
```
dmchoiboi: shouldn't we be passing in route into here, and getting a unique forwarder per route?

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -75,6 +93,11 @@ def load_streaming_forwarder(destination_path: Optional[str] = None) -> Streamin
     return get_streaming_forwarder_loader(destination_path).load(None, None)
 
 
+@lru_cache()
+def load_stream_passthrough_forwarder() -> PassthroughForwarder:
```
dmchoiboi: pass in destination_path?

model-engine/model_engine_server/inference/forwarding/forwarding.py
```diff
@@ -627,6 +627,84 @@ def endpoint(route: str) -> str:
...
+            if url.query:
+                target_url = f"{target_url}?{url.query}"
+
+            response = await aioclient.request(
+                method=request.method,
+                url=target_url,
+                data=await request.body() if request.method in ["POST", "PUT", "PATCH"] else None,
+                headers=headers,
+            )
+            yield response
```
dmchoiboi: i'm a little confused how this code works.

Shouldn't there be different handling based on whether the response is a sync or streaming response?

e.g. 
```
            if response.status != 200:
                yield await response.read()

            if is_streaming_response(response_headers):
                async for chunk in response.content.iter_chunks():
                    yield chunk[0]

            yield await response.read()
```

dmchoiboi: and 
```
def is_streaming_response(response_headers: dict) -> bool:
    content_type = response_headers.get("content-type", "")
    transfer_encoding = response_headers.get("transfer-encoding", "").lower()
    return (
        "text/event-stream" in content_type
        or "application/stream+json" in content_type
        or "application/octet-stream" in content_type
        or "multipart/x-mixed-replace" in content_type
        or "chunked" in transfer_encoding
    )
```

dmchoiboi: Also another code reference for a python forwarder proxy that i had to write up for a different use case: https://github.com/scaleapi/models/pull/13803/files#diff-744ca7f6f62a0bc1cf4f03973b55c61b81db11345fc40be8ef5e7965faee0ca7R229

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
MicahFulton: no since its passing through a request say `localhost:8000/mcp` forwarding to `localhost:5005/mcp` if we passed a destination that would just be forwarding to `localhost:5005/mcp/mcp`

model-engine/model_engine_server/inference/forwarding/forwarding.py
MicahFulton: updated.

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
MicahFulton: updated.

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
MicahFulton: updated.

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
MicahFulton: agreed.  Removed in latest.

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
MicahFulton: Added in latest.

model-engine/model_engine_server/inference/configs/service--http_forwarder.yaml
MicahFulton: Removed this config in latest in favor of filtering down to the props the PassthroughForwarder does accept

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -52,10 +57,40 @@ def get_streaming_forwarder_loader(
...
+
+    passthrough_forwarder_loader = LoadPassthroughForwarder(**config)
+    return passthrough_forwarder_loader
+
+
+def get_sync_passthrough_forwarder_loader(
+    destination_path: Optional[str] = None,
+) -> LoadPassthroughForwarder:
+    config = {}
+    stream_config = get_config().get("sync", {})
```
dmchoiboi: nit: sync_config

model-engine/model_engine_server/inference/forwarding/forwarding.py
```diff
@@ -627,6 +627,103 @@ def endpoint(route: str) -> str:
...
 
+@dataclass
+class PassthroughForwarder(ModelEngineSerializationMixin):
+    passthrough_endpoint: str
+
+    async def _make_request(
+        self, request: Any, aioclient: aiohttp.ClientSession
+    ) -> aiohttp.ClientResponse:
+        headers: dict[str, str] = dict(request.headers)
+        excluded_headers: set[str] = {
```
dmchoiboi, line 638: I believe you also have to exclude some headers from the response back. https://github.com/scaleapi/models/pull/13803/files#diff-c525735d4b3e0342c73386572f4eb59ea7af6556b5e69f5e6f49f44537a00928R40 is what i had to do at least

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -123,6 +170,33 @@ async def event_generator():
...
 
 
+async def passthrough_stream(
+    request: Request,
+    forwarder: PassthroughForwarder = Depends(get_stream_passthrough_forwarder_loader),
+    limiter: MultiprocessingConcurrencyLimiter = Depends(get_concurrency_limiter),
+):
+    with limiter:
+        response = forwarder.forward_stream(request)
+        headers, _ = await anext(response)
```
dmchoiboi: do you not need the status?

MicahFulton: whoops, just forgot it, good catch!

model-engine/model_engine_server/inference/forwarding/http_forwarder.py
```diff
@@ -123,6 +193,35 @@ async def event_generator():
...
+    with limiter:
+        response = forwarder.forward_stream(request)
+        headers, status = await anext(response)
+        headers = sanitize_response_headers(headers)
+
+        async def content_generator():
+            async for chunk in response:
+                yield chunk
+
+        return StreamingResponse(content_generator(), headers=headers, status_code=status)
```
dmchoiboi, line 210: this is a lot simpler than what i ended up with. 🤔  wonder why i ended up with such a mess

dmchoiboi (approved):
Thanks for incorporating feedback throughout the iterations!

PR #837, opened 2026-06-10
feat(api): retain reasoning content in v2 chat completions

olliestanley:
# Pull Request Summary

vLLM reasoning parsers return the model's reasoning (thinking) trace in a non-OpenAI-spec `reasoning` field on the chat completion message. The v2 chat completion DTOs validate upstream output against the generated OpenAI spec models (`common/types/gen/openai.py`), which silently drop the field — so reasoning traces never reach clients, and clients cannot round-trip reasoning back to chat templates that render prior-turn thinking (e.g. vLLM reasoning-parser deployments in agentic tool-calling loops). Today such deployments pay the decode cost of thinking while the trace is discarded in transit.

This PR extends the spec models via subclassing (the generated file is untouched, following the existing `ChatCompletionV2Request` precedent):

- **Sync responses & stream chunks**: messages/deltas gain `reasoning_content` (the LiteLLM/DeepSeek ecosystem convention), populated from either `reasoning` (vLLM) or `reasoning_content` upstream via a validation alias.
- **Requests**: assistant messages accept either `reasoning` or `reasoning_content` and forward the trace to the inference framework as `reasoning` — the field vLLM exposes to chat templates — since requests are forwarded with `model_dump(exclude_none=True)`.

No changes to the router or use cases — `ChatCompletionV2SyncResponse` / `ChatCompletionV2StreamSuccessChunk` keep their names (TypeAlias → subclass), so the existing `model_validate` call sites and the route's `response_model` pick up the extension automatically.

**Caveats:**
- Sync responses gain a `reasoning_content: null` key for non-reasoning models, consistent with the other nullable spec fields (`refusal`, `audio`, …) already serialized by FastAPI. Stream chunks and forwarded requests use `exclude_none=True` and are byte-identical for non-reasoning models.
- The pydantic-v1-based python client (`clients/python`) is not updated here (server DTO changes have historically not been mirrored there); raw HTTP / OpenAI SDK consumers get the field immediately.

## Test Plan and Usage Guide

Added `model-engine/tests/unit/common/test_chat_completion_dtos.py` (9 tests) covering: vLLM `reasoning` → `reasoning_content` mapping for sync and stream responses, acceptance of both input key conventions, request-side forwarding as `reasoning` (including an assistant tool-call turn, the main round-trip case), and unchanged serialization when no reasoning is present.

```
$ pytest tests/unit/common/test_chat_completion_dtos.py
9 passed
```

Also ran `black`, `isort`, `ruff`, and `mypy --config-file model-engine/mypy.ini` on the changed files (all clean).

Verified against a live deployment (vLLM 0.19.1, gemma-4 reasoning parser, GCP dev cluster): the raw endpoint returns `reasoning`, and `/v2/chat/completions` currently drops it; with these DTOs the same payload validates with `reasoning_content` populated.

<h3>Greptile Summary</h3>

This PR passes vLLM reasoning traces through the v2 chat completion layer by extending the generated OpenAI spec models via subclassing (the generated file is untouched). `ChatCompletionV2SyncResponse` and `ChatCompletionV2StreamSuccessChunk` are converted from TypeAliases to proper subclasses, with `reasoning_content` added to response messages and deltas; `ChatCompletionV2Request` gains an overridden `messages` field that routes assistant turns through a new `ChatCompletionRequestAssistantMessageWithReasoning` class that normalises `reasoning`/`reasoning_content` input to a canonical `reasoning` key for downstream forwarding.

- **Response path**: both sync responses and stream chunks accept `reasoning` (vLLM) or `reasoning_content` upstream and always surface `reasoning_content` to clients; non-reasoning stream chunks are byte-identical (serialized with `exclude_none=True`); sync responses gain a nullable `reasoning_content: null` key consistent with existing nullable spec fields.
- **Request path**: assistant messages accept either convention on input and forward to vLLM as `reasoning`, enabling agentic tool-calling loops to round-trip prior-turn thinking traces through chat templates.
- **Test coverage**: nine unit tests validate vLLM mapping, both input conventions, request-side forwarding, backward compatibility for non-reasoning models, and the `messages` min-length constraint that had to be explicitly re-declared on the field override.

Safe to merge — the change is additive, leaves non-reasoning model payloads byte-identical on stream/request paths, and all existing call sites pick up the extension automatically.

All three changed files are DTO/type-layer only with no changes to routing, use-case logic, or persistence. The field aliasing strategy is correct: validation aliases include the Python field names so both construction paths work; serialization always emits the canonical name (`reasoning_content` outbound, `reasoning` to vLLM). The `min_length=1` constraint is explicitly restated on the overridden `messages` field and confirmed by a dedicated test. Nine unit tests cover every combination that matters for production correctness.

No files require special attention.
</details>

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| model-engine/model_engine_server/common/dtos/llms/chat_completion.py | Core change: TypeAliases for sync response and stream chunk converted to proper subclasses; reasoning_content field added to response/delta models with validation aliases; messages field overridden with RootModel-based union that includes reasoning-aware assistant message; min_length=1 constraint correctly restated on the override. |
| model-engine/model_engine_server/common/pydantic_types.py | Minor: adds AliasChoices re-export alongside the existing re-exports; import order is slightly reshuffled. |
| model-engine/tests/unit/common/test_chat_completion_dtos.py | New test module with 9 tests covering: vLLM reasoning→reasoning_content mapping for sync and stream, both input key conventions, request-side forwarding as `reasoning`, tool-call round-trip, backward compatibility for non-reasoning models, and the messages min_length=1 constraint. |

</details>

```mermaid
sequenceDiagram
    participant Client
    participant Router as FastAPI Router
    participant UseCase as ChatCompletion UseCase
    participant vLLM as vLLM (upstream)

    Client->>Router: POST /v2/chat/completions
    Router->>Router: model_validate → ChatCompletionV2Request
    Note over Router: reasoning_content alias → reasoning field
    Router->>UseCase: execute(request)
    UseCase->>UseCase: "request.model_dump(exclude_none=True)"
    Note over UseCase: messages[i].reasoning forwarded as reasoning
    UseCase->>vLLM: "{messages: [{role:assistant, reasoning:...}]}"
    vLLM-->>UseCase: "{choices:[{message:{content:..., reasoning:trace}}]}"
    UseCase->>UseCase: ChatCompletionV2SyncResponse.model_validate(output)
    Note over UseCase: reasoning alias → reasoning_content field
    UseCase-->>Router: ChatCompletionV2SyncResponse
    Router-->>Client: "{choices:[{message:{content:..., reasoning_content:trace}}]}"
```
</details>

<sub>Reviews (4): Last reviewed commit: ["chore: retrigger CI (dropped CircleCI st..."](https://github.com/scaleapi/llm-engine/commit/26aba509623d5dda808935c58aebaa80a20ce34c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36736007)</sub>

### DIFF
--- a/model-engine/model_engine_server/common/dtos/llms/chat_completion.py
+++ b/model-engine/model_engine_server/common/dtos/llms/chat_completion.py
@@ -1,9 +1,19 @@
-from typing import Optional
+from typing import List, Optional, Union
 
 from model_engine_server.common.dtos.llms.completion import StreamError
 from model_engine_server.common.dtos.llms.vllm import VLLMChatCompletionAdditionalParams
-from model_engine_server.common.pydantic_types import BaseModel, Field
+from model_engine_server.common.pydantic_types import AliasChoices, BaseModel, Field, RootModel
 from model_engine_server.common.types.gen.openai import (
+    ChatCompletionRequestAssistantMessage,
+    ChatCompletionRequestDeveloperMessage,
+    ChatCompletionRequestFunctionMessage,
+    ChatCompletionRequestSystemMessage,
+    ChatCompletionRequestToolMessage,
+    ChatCompletionRequestUserMessage,
+    ChatCompletionResponseMessage,
+    ChatCompletionStreamResponseDelta,
+    Choice,
+    Choice1,
     CreateChatCompletionRequest,
     CreateChatCompletionResponse,
     CreateChatCompletionStreamResponse,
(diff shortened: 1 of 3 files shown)


billyang-scale (approved):
lgtm

olliestanley:
Addressed both review comments in 3f5ef36b:

**1. `min_length=1` on the `messages` override — kept, with a regression test.** The suggestion to drop it assumes the constraint is inherited from `CreateChatCompletionRequest.messages`, but pydantic v2 field overrides *replace* the parent `FieldInfo` rather than merging with it. Verified empirically: with the suggested change, `{"model": "m", "messages": []}` validates successfully (constraint silently lost); with the re-declaration it fails with `too_short`, matching the parent spec model. Added `test_request_requires_at_least_one_message` to lock this in, and a comment at the override site so the re-declaration doesn't read as accidental duplication.

**2. Production sync serialization — agreed, test added.** `test_sync_response_without_reasoning_serializes_stable_null` now asserts the default (FastAPI, no `exclude_none`) serialization emits a stable `reasoning_content: null` for non-reasoning models, alongside the pre-existing nullable spec fields (`refusal`, etc.) the route already emits. The `exclude_none` variant is kept since that's the contract for stream chunks and forwarded requests.

PR #874, opened 2026-08-27
fix(forwarder): stop the sync deadline capping async tasks at 1h (MLI-8587)

jay-tervala:
## What broke

`#868` gave the sync forwarder an explicit timeout to lift aiohttp's `total=300s` default off long non-streaming generations — a real fix. But the same `timeout_seconds` field also feeds the `requests.post` in `Forwarder.__call__`, which **only the Celery worker calls**, and `requests` has no default timeout. So the async path went from **unbounded to 3600s**. The same commit also derived the Celery hard `time_limit` from that field, giving async tasks a **second** 1-hour deadline that SIGKILLs the worker child.

Async endpoints then failed at exactly 60 minutes while the user container kept running and finished 30–90 minutes later, so callers discarded completed work. Because the hard limit fires before the HTTP timeout, the usual presentation is `FAILURE` with a **null traceback** — on `archie-qc-async`, **18 hard kills to 3 `ReadTimeout`s** over five days.

Reported in MLI-8587 (blocking terminal-bench-3). The endpoint crossed over at `2026-08-22T15:50:02Z`, when its next rebuild swept in the newer forwarder image.

## What this changes

- **The async forwarder no longer inherits the sync deadline.** It defaults to the queue's visibility window instead. The grace is subtracted from it so the hard limit lands *on* the visibility boundary rather than past it, where the message would already have been redelivered to another worker. `86100s` HTTP + `300s` grace = `86400s` = `DEFAULT_TASK_VISIBILITY_SECONDS`.
- **The Celery hard limit is held above the HTTP timeout**, so the forwarder's own `ReadTimeout` raises first and a deadline hit arrives with a traceback instead of as a bare kill.
- **`GetAsyncTaskV1Response` reports `TaskStatus.TIMEOUT`** for a deadline rather than `FAILURE`. A timed-out task leaves the container running, so the result may still land; callers previously could not tell that from a crash, because the API stringifies the exception and `str(TimeLimitExceeded(3600))` is just `"3600"`. Every non-deadline failure stays `FAILURE`.

The config file and `--set` stay authoritative over the new default.

## Blast radius this addresses

515 async endpoints carry a `celery-forwarder`; only 10 had picked up the affected image, because the forwarder tag is frozen at endpoint **build** time and each endpoint converts on its next rebuild. The remaining ~505 would have acquired the cap one at a time over the following weeks.

## Risk

Sync and streaming endpoints are untouched. `Forwarder.__call__` has exactly **one** caller in the tree (`celery_forwarder.py`); `http_forwarder.py` uses `await forwarder.forward(...)`, the aiohttp path. `#868`'s 300s → 3600s improvement for sync is unchanged.

## Tests

- `test_celery_forwarder_worker.py` — the existing wall-clock-limit test now asserts the ordering invariant (hard limit strictly above the HTTP timeout); new tests cover the async default not inheriting the sync deadline, the whole budget fitting inside the visibility window, and config/`--set` still winning.
- `test_celery_task_queue_gateway.py` — deadline exceptions (celery hard limit, `ReadTimeout`, `ConnectTimeout`) map to `TIMEOUT`; `RuntimeError`/`ValueError`/`MemoryError` stay `FAILURE`.
- `model-engine/tests/unit`: **846 passed, 9 skipped**. One pre-existing failure, `test_image_cache_gateway::test_create_or_update_image_cache`, also fails on a clean `origin/main` tree — unrelated to this change.
- black / isort / ruff / mypy clean on the touched files.

## Deliberately out of scope

- **Per-endpoint tunability.** There is still no way to set the async deadline for a single endpoint: the chart only passes `predict_route` and `healthcheck_route` through to the celery-forwarder, and adding a knob needs the endpoint builder and resource types too. Worth a follow-up; kept out so this can land as the unblock.
- **Per-request `timeout_seconds` on the async predict body** (MLI-8587 ask #3). The forwarder deadline is fixed at load time and the Celery limit at task-decoration time, so this needs changes in both components.
- **`result=str(res.result)` discards the exception type for *all* async failures**, not just timeouts. `TaskStatus.TIMEOUT` covers the reported case; changing the `result` shape is a wider contract change.
- The committed OpenAPI specs are **already stale on `main`** (regenerating pulls in 8 added / 8 removed schemas and 4 changed paths from earlier work). Only the `TaskStatus` enum is updated here rather than importing that drift.

Related: **MLI-8206** is the ticket `#868`'s forwarder-timeout sub-commit was written for — its fix is this regression's cause, and it is still open.

<h3>Greptile Summary</h3>

The PR separates asynchronous inference deadlines from the synchronous forwarder timeout and reports deadline failures distinctly.
- Defaults the async HTTP timeout to 86,100 seconds and caps overrides so the 300-second Celery grace remains within the 24-hour visibility window.
- Maps recognized deadline exceptions to the new `TIMEOUT` task status and treats that status as terminal during batch polling.
- Updates the public schemas and adds focused timeout, visibility-budget, status-mapping, and polling tests.

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| model-engine/model_engine_server/inference/forwarding/celery_forwarder.py | Separates the async timeout default from the sync deadline, caps configured overrides, and bounds the Celery hard limit by queue visibility. |
| model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py | Maps recognized timeout exception types from failed Celery results to the new public TIMEOUT status. |
| model-engine/model_engine_server/common/dtos/tasks.py | Adds TIMEOUT to the shared task-status contract. |
| model-engine/model_engine_server/infra/services/live_batch_job_orchestration_service.py | Treats TIMEOUT as terminal so batch polling completes instead of repeatedly reading the result backend. |
| model-engine/specs/openapi-3.0.json | Adds TIMEOUT to the OpenAPI 3.0 TaskStatus enum. |
| model-engine/specs/openapi.json | Adds TIMEOUT to the committed OpenAPI TaskStatus enum. |

</details>

```mermaid
sequenceDiagram
    participant Client
    participant Broker as Celery Broker
    participant Worker as Async Forwarder
    participant Model as Model Service
    participant Backend as Result Backend
    Client->>Broker: Submit async inference
    Broker->>Worker: Deliver task
    Worker->>Model: HTTP request (86,100s maximum)
    alt Model responds before deadline
        Model-->>Worker: Result
        Worker->>Backend: Store SUCCESS
    else HTTP deadline expires
        Worker->>Backend: Store timeout exception
        Backend-->>Client: TIMEOUT
    else Celery hard limit reached at visibility boundary
        Worker--xWorker: Hard termination
        Backend-->>Client: TIMEOUT
    end
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(batch): treat TIMEOUT as terminal in..."](https://github.com/scaleapi/llm-engine/commit/0e2b214fe029bcfb32d86c6f60a67263f007de8f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57583457)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [Inference runtime](https://app.greptile.com/scale-ai/-/custom-context/knowledge-base/scaleapi/llm-engine/-/docs/inference-runtime.md)
- Knowledge Base — [Asynchronous inference](https://app.greptile.com/scale-ai/-/custom-context/knowledge-base/scaleapi/llm-engine/-/docs/asynchronous-inference.md)
- Knowledge Base — [Batch-job orchestration and image builds](https://app.greptile.com/scale-ai/-/custom-context/knowledge-base/scaleapi/llm-engine/-/docs/batch-job-orchestration.md)
- Knowledge Base — [Service composition and background workers](https://app.greptile.com/scale-ai/-/custom-context/knowledge-base/scaleapi/llm-engine/-/docs/service-composition-and-workers.md)
</details>

### DIFF
--- a/model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
+++ b/model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
@@ -26,6 +26,18 @@ except ImportError:
     ServiceBusError = None  # type: ignore[assignment,misc]
 
 logger = make_logger(logger_name())
+
+# Matched by name, not isinstance: the result backend rebuilds the exception from the stored
+# exc_type/exc_module, and synthesizes a stand-in class when that module is not importable here.
+TIMEOUT_EXCEPTION_NAMES = frozenset(
+    {
+        "TimeLimitExceeded",  # celery/billiard hard limit; the worker child was killed
+        "SoftTimeLimitExceeded",
+        "Timeout",  # requests, and the base of the two below
+        "ConnectTimeout",
+        "ReadTimeout",
+    }
+)
 _cloud_provider = infra_config().cloud_provider
 backend_protocol = (
     "abs" if _cloud_provider == "azure" else ("redis" if _cloud_provider == "gcp" else "s3")
@@ -399,9 +411,12 @@ class CeleryTaskQueueGateway(TaskQueueGateway):
             )
 
         elif response_state == "FAILURE":
+            # A deadline hit leaves the user container still running, so it is reported apart from
+            # a crash: the work may yet land and the caller can keep polling rather than discard it.
+            timed_out = type(res.result).__name__ in TIMEOUT_EXCEPTION_NAMES
             return GetAsyncTaskV1Response(
                 task_id=task_id,
-                status=TaskStatus.FAILURE,
+                status=TaskStatus.TIMEOUT if timed_out else TaskStatus.FAILURE,
                 result=str(res.result) if res.result is not None else None,
                 traceback=res.traceback,
                 status_code=None,  # probably
--- a/model-engine/model_engine_server/common/dtos/tasks.py
+++ b/model-engine/model_engine_server/common/dtos/tasks.py
@@ -22,6 +22,7 @@ class TaskStatus(str, Enum):
     STARTED = "STARTED"
     SUCCESS = "SUCCESS"
     FAILURE = "FAILURE"
+    TIMEOUT = "TIMEOUT"
     UNDEFINED = "UNDEFINED"
 
 
--- a/model-engine/model_engine_server/inference/forwarding/celery_forwarder.py
+++ b/model-engine/model_engine_server/inference/forwarding/celery_forwarder.py
@@ -48,6 +48,17 @@ logger = make_logger(logger_name())
 
 tracing_gateway = get_tracing_gateway()
 
+# The Celery hard limit SIGKILLs the worker child, which discards the traceback. Held above the
+# forwarder's own HTTP timeout so that timeout raises first and a deadline hit stays diagnosable.
+CELERY_TIME_LIMIT_GRACE_SECONDS: float = 300
+
+# An async task is bounded by how long its queue will wait for an ack, not by the sync request
+# deadline. The grace is subtracted so the hard limit lands on the visibility boundary rather than
+# past it, where the message has already been redelivered to another worker.
+DEFAULT_ASYNC_TIMEOUT_SECONDS: float = (
+    DEFAULT_TASK_VISIBILITY_SECONDS - CELERY_TIME_LIMIT_GRACE_SECONDS
+)
+
 
 class ErrorResponse(TypedDict):
     """The response payload for any inference request that encountered an error."""
@@ -103,7 +114,7 @@ def create_celery_service(
     monitoring_metrics_gateway = DatadogInferenceMonitoringMetricsGateway()
     # requests treats its timeout as socket inactivity. The Celery limit supplies
     # the wall-clock bound when an upstream response keeps trickling bytes.
-    task_time_limit = forwarder.timeout_seconds
+    task_time_limit = forwarder.timeout_seconds + CELERY_TIME_LIMIT_GRACE_SECONDS
 
     class ErrorHandlingTask(Task):
         """Sets a 'custom' field with error in the Task response for FAILURE.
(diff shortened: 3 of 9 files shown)


lorenzo-norcini-scale (dismissed):

model-engine/model_engine_server/infra/gateways/celery_task_queue_gateway.py
```diff
@@ -399,9 +411,12 @@ def get_task(self, task_id: str) -> GetAsyncTaskV1Response:
             )
 
         elif response_state == "FAILURE":
+            # A deadline hit leaves the user container still running, so it is reported apart from
+            # a crash: the work may yet land and the caller can keep polling rather than discard it.
+            timed_out = type(res.result).__name__ in TIMEOUT_EXCEPTION_NAMES
             return GetAsyncTaskV1Response(
                 task_id=task_id,
-                status=TaskStatus.FAILURE,
+                status=TaskStatus.TIMEOUT if timed_out else TaskStatus.FAILURE,
```
lorenzo-norcini-scale, line 419: TIMEOUT is terminal here, but the batch poller only stops on SUCCESS or FAILURE. Should we add TIMEOUT to its terminal states?

lorenzo-norcini-scale: https://github.com/scaleapi/llm-engine/blob/b3b18ce44564ef4afcbb5a03dfc8ddba7af40374/model-engine/model_engine_server/infra/services/live_batch_job_orchestration_service.py#L328-L339

lorenzo-norcini-scale: Not even sure if this path is used tbh

lorenzo-norcini-scale (approved):

